PR #46, opened 2015-10-07
added notes for MutationObserver differences

### LINKED ISSUE #37: Why the slight differences from MutationObserver?
There are a number of slight differences in design from MutationObserver:

- IntersectionObserverInit is in the constructor, whereas for MutationObserver it is in the observe method.
- IntersectionObserver has an unobserve() method; MutationObserver does not.

These changes are probably well-motivated, but since I'm not too familiar with the design space, I find the mismatch surprising. If they are intentional, it would be nice to add a non-normative note explaining them. It might be the case that MutationObserver has flaws and this is a slightly better iteration.

mpb:
Fixes #37


index.bs
```diff
@@ -99,9 +99,23 @@ spec:dom-ls; type:interface; text:Document
...
+    {{Element}} (or the top-level document's viewport).
+
+    A <dfn>registered intersection observer</dfn> consists of an observer (an
+    {{IntersectionObserver}} object) and options (an
+    {{IntersectionObserverInit}} dictionary).
+
+    Note: In {{MutationObserver}}, the {{MutationObserverInit}} options are
+    passed to {{MutationObserver/observe()}} while in {{IntersectionObserver}}
+    they are passed to the constructor. This is because for MutationObserver,
+    each {{Node}} being observerd could have a different set of attributes to
```
domenic: typo observerd

index.bs
```diff
@@ -99,9 +99,23 @@ spec:dom-ls; type:interface; text:Document
...
+    filter. For IntersectionObserver, having different {{root}},
+    {{rootMargin}} or {{threshold}} values for each <i>target</i> seems to
+    introduce more complexity without solving additional use-cases. A
+    per-{{observe()}} init could be provided in the future if V2 introduces
```
domenic: "A per-observe() init" => "Per-observe() options"

domenic:
LGTM with nits. Thanks!

BTW if you use "Fixes" instead of "Addresses" in your commit message, GitHub will automatically close the issue that you reference. https://help.github.com/articles/closing-issues-via-commit-messages/

PR #47, opened 2015-10-07
Processing model re-write based on MutationObserver.

### LINKED ISSUE #35: Use of undefined variable "Document" throughout
All the processing model algorithms make reference to _Document_. But this is not defined. I assume it's meant to be a Document of some sort. But which one? From which window? This can be especially bad when e.g. code in one frame uses the IntersectionObserver constructor from another frame to observe an Element from a third frame.

### LINKED ISSUE #43: [[RegisteredIntersectionObservers]] setup is broken
This might be related to #35.

[[RegisteredIntersectionObservers]] is defined on Document and used in https://rawgit.com/slightlyoff/IntersectionObserver/master/index.html#3-3-event-loop, but nobody ever adds to it or removes from it.

In contrast, another [[RegisteredIntersectionObservers]] is defined on Element and modified by IntersectionObserver's methods, but nobody ever reads from it.

Again it might be worth consulting what Mutation Observers do here.

mpb:
Fixes #35 
Fixes #43

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -13,14 +13,10 @@ Repository: slightlyoff/IntersectionObserver
 <pre class="anchors">
 urlPrefix: http://www.w3.org/TR/hr-time/; type: typedef; text: DOMHighResTimeStamp
 url: https://drafts.csswg.org/css-values/#lengths; type: attribute; text: length
-urlPrefix: https://html.spec.whatwg.org/multipage/browsers.html; type: dfn; text: unit of related similar-origin browsing contexts
 url: http://w3c.github.io/requestidlecallback/#dfn-list-of-idle-request-callbacks; type: dfn; text: list of idle request callbacks
 url: https://html.spec.whatwg.org/multipage/infrastructure.html#dfn-callback-this-value; type: dfn; text: callback this value
+urlPrefix: https://html.spec.whatwg.org/multipage/browsers.html; type: dfn; text: unit of related similar-origin browsing contexts
 urlPrefix: https://html.spec.whatwg.org/multipage/webappapis.html; type: dfn; text: report the exception
-urlPrefix: https://drafts.fxtf.org/css-masking-1/; type: dfn; text: the clip path
-url: https://dom.spec.whatwg.org/#concept-node-remove-ext; type: dfn; text: removing steps
-url: https://dom.spec.whatwg.org/#concept-tree-inclusive-descendant; type: dfn; text: inclusive descendants
-url: https://dom.spec.whatwg.org/#concept-tree-order; type: dfn; text: tree order
 urlPrefix: https://html.spec.whatwg.org/multipage/webappapis.html; type: dfn; text: event loop
 </pre>
 
@@ -415,9 +411,8 @@ Observations Steps</h4>
                         <i>target</i> and <i>observer</i>'s internal
                         {{[[root]]}} slot.</li>
 
-                        Issue: TBD: Do the <i>clip rect</i>s include <a>the clip
-                        path</a> and other clipping properties or just overflow
-                        clipping?
+                        Issue: TBD: Do the <i>clip rect</i>s include 'clip-path'
+                        and other clipping properties or just overflow clipping?
                     <li>Let <i>newIntersectionExceeds</i> be <i>true</i> if
                         <i>intersectionRect</i>'s area is greater than or equal
                         to <i>observer</i>'s internal {{[[threshold]]}} slot,
--- a/index.bs
+++ b/index.bs
@@ -419,7 +419,7 @@ Observations Steps</h4>
                     <li>Let <i>oldIntersectionExceeds</i> be <i>true</i> if
                         <i>target</i>'s internal
                         {{[[PreviousIntersectionRect]]}} slot is not null and
-                        has an area that is great than or equal to
+                        has an area that is greater than or equal to
                         <i>observer</i>'s internal {{[[threshold]]}} slot and
                         <i>false</i> otherwise.</li>
                     <li>If <i>newIntersectionExceeds</i> does not equal
--- a/index.bs
+++ b/index.bs
@@ -11,9 +11,14 @@ Repository: slightlyoff/IntersectionObserver
 </pre>
 
 <pre class="anchors">
-text: DOMHighResTimeStamp; type: typedef; urlPrefix: http://www.w3.org/TR/hr-time/
-text: margin; type: attribute; url: http://www.w3.org/TR/CSS21/box.html#propdef-margin
-text: length; type: attribute; url: http://www.w3.org/TR/css3-values/#lengths
+urlPrefix: http://www.w3.org/TR/hr-time/; type: typedef; text: DOMHighResTimeStamp
+url: http://www.w3.org/TR/CSS21/box.html#propdef-margin; type: attribute; text: margin
+url: http://www.w3.org/TR/css3-values/#lengths; type: attribute; text: length
+urlPrefix: http://www.w3.org/TR/html5/browsers.html; type: dfn; text: unit of related similar-origin browsing contexts
+url:http://w3c.github.io/requestidlecallback/#dfn-list-of-idle-request-callbacks; type: dfn; text: list of idle request callbacks
+url:http://www.w3.org/TR/WebIDL-1/#dfn-callback-this-value; type: dfn; text: callback this value
+urlPrefix:http://www.w3.org/TR/html51/webappapis.html; type: dfn; text: report the exception
+url:http://www.w3.org/TR/2014/WD-css-masking-1-20140213/#the-clip-path; type: dfn; text: clip path
 </pre>
 
 <pre class="link-defaults">
@@ -149,8 +154,7 @@ The IntersectionObserver interface</h3>
                 <li> Set <i>this</i>'s internal {{[[threshold]]}} slot to
                     <i>options</i>.{{threshold}}</li>
                 <li>Append <i>this</i> to the
-                    <a href="http://www.w3.org/TR/html5/browsers.html#unit-of-related-similar-origin-browsing-contexts">
-                    unit of related similar-origin browsing contexts</a>'s
+                    <a>unit of related similar-origin browsing contexts</a>'s
                     internal <a>\[[IntersectionObservers]]</a> slot.</li>
                 <li>return <i>this</i>.</li>
             </ol>
@@ -304,9 +308,7 @@ The IntersectionObserverInit dictionary</h3>
 <h3 id='3-1-defines'>Internal Slot Definitions</h2>
 
 <h4 id='3-1-1-browsing-contexts'>Browsing Contexts</h4>
-    Each
-    <a href="http://www.w3.org/TR/html5/browsers.html#unit-of-related-similar-origin-browsing-contexts">
-    unit of related similar-origin browsing contexts</a> has an internal
+    Each <a>unit of related similar-origin browsing contexts</a> has an internal
     <dfn>\[[IntersectionObserverTaskQueued]]</dfn> slot which is initialized to
     false and an internal <dfn>\[[IntersectionObservers]]</dfn> slot which is
     initialized to an empty list.
@@ -341,8 +343,7 @@ Task</h4>
             true.</li>
         <li>Post a task to <a>notify intersection observers</a>, or
             enqueue a task to <a>notify intersection observers</a> in the
-            <a href="https://w3c.github.io/requestidlecallback/#dfn-list-of-idle-request-callbacks">
-            list of idle request callbacks</a> with an appropriate
+            <a>list of idle request callbacks</a> with an appropriate
             <i>timeout</i>.
 
             Issue: Should we define an appropriate <i>timeout</i>?
@@ -355,8 +356,7 @@ Task</h4>
         <li>Set the internal <a>\[[IntersectionObserverTaskQueued]]</a> slot to
             false.</li>
         <li>Let <i>notify list</i> be a copy of
-            <a href="http://www.w3.org/TR/html5/browsers.html#unit-of-related-similar-origin-browsing-contexts">
-            unit of related similar-origin browsing contexts</a>'s internal
+            <a>unit of related similar-origin browsing contexts</a>'s internal
             <a>\[[IntersectionObservers]]</a> slot.</li>
         <li>For each {{IntersectionObserver}} object <i>observer</i> in
             <i>notify list</i>, run these steps:</li>
@@ -370,10 +370,8 @@ Task</h4>
                 </li>
                 <li>Invoke <i>callback</i> with <i>queue</i> as the first
                 argument and <i>observer</i> as the second argument and
-                <a href="http://www.w3.org/TR/WebIDL-1/#dfn-callback-this-value">
-                callback this value</a>. If this throws an exception,
-                <a href="http://www.w3.org/TR/html51/webappapis.html#report-the-exception">
-                report the exception</a>.</li>
+                <a>callback this value</a>. If this throws an exception,
+                <a>report the exception</a>.</li>
             </ol>
     </ol>
 
@@ -395,9 +393,8 @@ Observations Steps</h4>
     steps:
         <ul>
             <li>For each <i>observer</i> in the
-                <a href="http://www.w3.org/TR/html5/browsers.html#unit-of-related-similar-origin-browsing-contexts">
-                unit of related similar-origin browsing contexts</a>'s internal
-                <a>\[[IntersectionObservers]]</a> slot</li>
+                <a>unit of related similar-origin browsing contexts</a>'s
+                internal <a>\[[IntersectionObservers]]</a> slot</li>
             <ul>
                 <li>For each <i>target</i> in <i>observer</i>'s internal
                 {{[[ObservationTargets]]}} slot</li>
@@ -429,8 +426,7 @@ Observations Steps</h4>
                         </ul>
 
                         Issue: TBD: Does <i>clipRect</i> include
-                        <a href="http://www.w3.org/TR/2014/WD-css-masking-1-20140213/#the-clip-path">
-                        clip path</a> and other clipping properties or just
+                        <a>clip path</a> and other clipping properties or just
                         overflow clipping?
                     </li>
                     <li>Assign <i>intersctionRect</i> to <i>target</i>'s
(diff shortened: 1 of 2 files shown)


mpb:
@domenic, PTAL and see if this addresses your concerns with the processing model. 
(It's kind of a large change, so I may have missed something.)
Thanks!

index.bs
```diff
@@ -8,14 +8,17 @@ Editor: Michael Blain, Google, mpb@google.com
...
 Repository: slightlyoff/IntersectionObserver
-Mailing List: public-web-perf@w3c.org
-Mailing List Archives: http://lists.w3.org/Archives/Public/public-web-perf/
 </pre>
 
 <pre class="anchors">
-text: DOMHighResTimeStamp; type: typedef; urlPrefix: http://www.w3.org/TR/hr-time/
-text: margin; type: attribute; url: http://www.w3.org/TR/CSS21/box.html#propdef-margin
-text: length; type: attribute; url: http://www.w3.org/TR/css3-values/#lengths
+urlPrefix: http://www.w3.org/TR/hr-time/; type: typedef; text: DOMHighResTimeStamp
```
domenic: Drive-by comment: prefer EDs to TRs. Also Web IDL 1 is dead; link to http://heycam.github.io/webidl/ instead.

index.bs
```diff
@@ -92,13 +101,16 @@ spec:dom-ls; type:interface; text:Document
...
+The IntersectionObserverCallback</h3>
+    <pre class="idl">
+        callback IntersectionObserverCallback = void (sequence&lt;IntersectionObserverEntry> entries, IntersectionObserver observer)
+    </pre>
+    This callback will be invoked when there are changes to <i>target</i>'s
+    intersection with <i>root</i>, as per the <a>processing model</a>.
 
-<h3 id='2-2-intersection-observer-interface'>The IntersectionObserver interface</h3>
+<h3 id='2-2-intersection-observer-interface'>
+The IntersectionObserver interface</h3>
```
domenic: The "registered intersection observer" definition seems redundant. Probably best to get rid of it, and in the two places it is referenced, say, "this IntersectionObserver object"

index.bs
```diff
@@ -128,7 +140,25 @@ This callback will be invoked when there are changes to <i>target</i>'s intersec
...
+                    <i>callback</i>.</li>
+                <li> Set <i>this</i>'s internal {{[[root]]}} slot to
+                    <i>options</i>.{{root}}</li>
+                <li> Set <i>this</i>'s internal {{[[rootMargin]]}} slot to
+                    <i>options</i>.{{rootMargin}}</li>
+                <li> Set <i>this</i>'s internal {{[[threshold]]}} slot to
+                    <i>options</i>.{{threshold}}</li>
+                <li>Append <i>this</i> to the
+                    <a>unit of related similar-origin browsing contexts</a>'s
+                    internal <a>\[[IntersectionObservers]]</a> slot.</li>
```
domenic: Styling seems off for [[IntersectionObservers]]

index.bs
```diff
@@ -128,7 +140,25 @@ This callback will be invoked when there are changes to <i>target</i>'s intersec
...
+                <li>return <i>this</i>.</li>
```
domenic: Capitalize return

index.bs
```diff
@@ -128,7 +140,25 @@ This callback will be invoked when there are changes to <i>target</i>'s intersec
     </pre>
 
     <div dfn-type="method" dfn-for="IntersectionObserver">
-        : <dfn>observe()</dfn>
+        : <dfn>Constructor(callback, options)</dfn>
```
domenic: There's no established convention I know of for this sort of thing but I personally like the heading to be `new IntersectionObserver(callback, options)`.

index.bs
```diff
@@ -128,7 +140,25 @@ This callback will be invoked when there are changes to <i>target</i>'s intersec
...
+                <li> Set <i>this</i>'s internal {{[[callback]]}} slot to
+                    <i>callback</i>.</li>
+                <li> Set <i>this</i>'s internal {{[[root]]}} slot to
+                    <i>options</i>.{{root}}</li>
+                <li> Set <i>this</i>'s internal {{[[rootMargin]]}} slot to
+                    <i>options</i>.{{rootMargin}}</li>
+                <li> Set <i>this</i>'s internal {{[[threshold]]}} slot to
+                    <i>options</i>.{{threshold}}</li>
+                <li>Append <i>this</i> to the
+                    <a>unit of related similar-origin browsing contexts</a>'s
```
domenic: FYI filed https://github.com/whatwg/dom/issues/88 to figure out if this should be more precise. No need to block this PR on such deliberations though.

index.bs
domenic: Oh, or just _this_, which I see you've used. (There is no good agreement on how to refer to "this", so it seems fine to say _this_ since it's clear.)

index.bs
domenic: In general sometimes it seems slots are code-styled, sometimes sans-serif. Worth doing a once-over.

index.bs
```diff
@@ -294,79 +325,76 @@ This callback will be invoked when there are changes to <i>target</i>'s intersec
...
     which are initialized to empty lists and internal
+    <dfn attribute for=IntersectionObserver>\[[callback]]</dfn>,
     <dfn attribute for=IntersectionObserver>\[[root]]</dfn>,
     <dfn attribute for=IntersectionObserver>\[[rootMargin]]</dfn> and
     <dfn attribute for=IntersectionObserver>\[[threshold]]</dfn> slots, which
-    are initialized to the corresponding attributes of the {{options}} parameter
-    passed to <a>IntersectionObserver(callback, options)</a>.
+    are initialized by <a>IntersectionObserver(callback, options)</a>.
 
 <h3 id='3-2-algorithms'>Algorithms</h2>
```
domenic: All of these algorithms are a bit too implicit on what they are operating. I would state them as e.g. "To **notify intersection observers**, given a unit of related similar-origin browser contexts _unit_, run these steps: ... If _unit_'s internal  [[IntersectionObserverTaskQueued]] slot is..."

index.bs
```diff
@@ -294,79 +325,76 @@ This callback will be invoked when there are changes to <i>target</i>'s intersec
...
-<h4 id='3-2-3-queue-intersection-observer-entry'>
-<dfn>Queue an IntersectionObserverEntry</dfn>
-</h4>
-    To <a>queue an IntersectionObserverEntry</a> for <i>observer</i>, run
+<h4 id='3-2-3-queue-intersection-observer-entry'>Queue an
+IntersectionObserverEntry</dfn></h4>
+    To <dfn>queue an IntersectionObserverEntry</dfn> for <i>observer</i>, run
     these steps:
     <ol>
         <li>Construct an {{IntersectionObserverEntry}}.</li>
```
domenic: So, unless I am missing something, at no point do you actually set the values of the IntersectionObserverEntry's properties. That seems pretty bad...

index.bs
```diff
@@ -398,8 +426,7 @@ This callback will be invoked when there are changes to <i>target</i>'s intersec
                         </ul>
 
                         Issue: TBD: Does <i>clipRect</i> include
-                        <a href="http://www.w3.org/TR/2014/WD-css-masking-1-20140213/#the-clip-path">
-                        clip path</a> and other clipping properties or just
+                        <a>clip path</a> and other clipping properties or just
                         overflow clipping?
                     </li>
                     <li>Assign <i>intersctionRect</i> to <i>target</i>'s
```
domenic: typo: intersctionRect

index.bs
```diff
@@ -407,4 +434,30 @@ This callback will be invoked when there are changes to <i>target</i>'s intersec
                 </ol>
             </ul>
         </ul>
+
+<h3 id='3-3-external-spec-integrations'>External Spec Integrations</h3>
+
+<h4 id='3-3-1-event-loop'>HTML Processing Model: Event Loop</h4>
+    An <a>Intersection Observer</a> processing step should take place AFTER the
+    layout and rendering steps have been performed in the
+    <a href="https://html.spec.whatwg.org/multipage/webappapis.html#processing-model-8">
```
domenic: Given that you're linking to the latest HTML spec here, probably best to change the link defaults in the header to also do so. Otherwise you are linking to both the outdated fork and the latest HTML spec in the same document and it's pretty confusing.

index.bs
```diff
@@ -407,4 +434,30 @@ This callback will be invoked when there are changes to <i>target</i>'s intersec
...
+    update the rendering</a> event loop in the HTML Processing Model.
+
+    This step is:
+    <ol start="11">
+        <li><a>Run the update intersection observations steps</a> for the
+        <a href="https://html.spec.whatwg.org/multipage/webappapis.html#event-loop">
+        event loop</a> in question, passing in <i>now</i> as the timestamp.</li>
+    </ol>
+
+<h4 id='3-3-2-remove-node'>DOM4: Remove a node</h4>
```
domenic: A monkey-patch on DOM is actually not necessary. You can just define [removing steps](https://dom.spec.whatwg.org/#concept-node-remove-ext) for all element nodes.

domenic:
Looking very good over all. This model is much easier for me to understand at least. I hope it doesn't mismatch the implementation too much. But it definitely clarifies a lot of things. Thanks for taking the time.

index.bs
mpb: Sure, will update the webidl link.
When I look for these specs, the TRs seem to be what turns up. Were there ones in particular you had in mind?

index.bs
mpb: IntersectionObservers was the one private slot I couldn't figure out how to add as an IDL and had to do it as a DFN. Because it's associated with a 'browsing context' which isn't really an object, I couldn't figure out how to make it a property. So that's why I always used an \<a> to link to it instead of a {{ }}. Any suggestions?

index.bs
mpb: Done.

index.bs
mpb: Done.

index.bs
mpb: Done.

index.bs
mpb: Done.

index.bs
domenic: TRs have an unfortunate amount of google-juice, it's true. But they are outdated and should not be used as a basis for implementations. So, everywhere I see /TR/ in this list, it's better to navigate to that URL and click the "editor's draft" link, and use that instead.

index.bs
domenic: Mmm, no real ideas then; maybe @TabAtkins has some Bikeshed tips for us.

tabatkins: If it's not an object (not an IDL interface), it doesn't have private slots, it just has, you know, *stuff*.  Literally all of it is private if it's not an object.  It's just a concept.  Just use `<dfn for="browsing context">whatever</dfn>`, then `<a for="browsing context">whatever</a>`.

index.bs
tabatkins: Note, tho, a constructor isn't a method. You need `<dfn constructor lt="IntersectionObserver(callback, options)">...</dfn>`.

mpb: What goes in the ...?
```<dfn constructor lt="IntersectionObserver(callback, options)">IntersectionObserver(callback, options)</dfn>```?

File: index.bs [hunk 12]
```diff
@@ -92,13 +101,16 @@ spec:dom-ls; type:interface; text:Document
     the visibility and position of DOM elements relative to a root element or
     the top level document's viewport.
 
-<h3 id='2-1-intersection-observer-callback'>The IntersectionObserverCallback</h3>
-<pre class="idl">
-  callback IntersectionObserverCallback = void (sequence&lt;IntersectionObserverEntry> entries, IntersectionObserver observer)
-</pre>
-This callback will be invoked when there are changes to <i>target</i>'s intersection with <i>root</i>, as per the <a>processing model</a>.
+<h3 id='2-1-intersection-observer-callback'>
```
tabatkins: Don't put section numbers in the ID. If you ever move/add/delete sections, you'll either have to change the ID (bad) or have the number be wrong (bad).

index.bs
tabatkins: Whatever you want to display. The `lt` overrides the linking text, so you can put whatever you want as the actual content.  What Domenic suggested is fine.

mpb: Done.

index.bs
mpb: I actually imagine that the whole 3.3 section goes away, and these steps would literally get patched into the Webapps API spec. Because of that, I wanted to skip pulling the references up into the main <pre class=anchors> block so I don't forget to clean it up. Is that reasonable?

index.bs
mpb: Done.

index.bs
mpb: Done.

index.bs
mpb: Done.

index.bs
mpb: Done.

index.bs
domenic: Right, that makes sense, and I agree. All I'm saying is that given that you're linking to https://html.spec.whatwg.org/multipage/ here, you should change the link defaults (at the top of the .bs file) from http://www.w3.org/TR/html5/ to https://html.spec.whatwg.org/multipage/.

index.bs
```diff
@@ -270,103 +299,102 @@ This callback will be invoked when there are changes to <i>target</i>'s intersec
...
-        <li>Append <i>observer</i> to <i>Document</i>'s internal
-            {{[[PendingIntersectionObserverCallbacks]]}} slot.</li>
-        <li>Post a task to <a>invoke an IntersectionObserverCallback</a>, or
-            enqueue a task to <a>invoke an IntersectionObserverCallback</a> in
-            <i>Document</i>'s
-            <a href="https://w3c.github.io/requestidlecallback/#dfn-list-of-idle-request-callbacks">
-            list of idle request callbacks</a> with an appropriate
+        <li>If <i>unit</i>'s <a>IntersectionObserverTaskQueued</a> flag is set
+            to true, return.</li>
+        <li>Set <i>unit<\i>'s <a>IntersectionObserverTaskQueued</a> flag to
```
domenic: typo \ vs. /

mpb: Done.

index.bs
mpb: Done.
hr-time and css3-box I could only find as TRs. 
for css-value and css-masking-1 I found EDs.

tabatkins: Don't link to css3-box, it's an obsolete spec and not worth anything.  It looks like that's not actually used, tho? Linking to 2.1 for margin is correct.

index.bs
```diff
@@ -8,14 +8,20 @@ Editor: Michael Blain, Google, mpb@google.com
...
-Mailing List: public-web-perf@w3c.org
-Mailing List Archives: http://lists.w3.org/Archives/Public/public-web-perf/
 </pre>
 
 <pre class="anchors">
-text: DOMHighResTimeStamp; type: typedef; urlPrefix: http://www.w3.org/TR/hr-time/
-text: margin; type: attribute; url: http://www.w3.org/TR/CSS21/box.html#propdef-margin
-text: length; type: attribute; url: http://www.w3.org/TR/css3-values/#lengths
+urlPrefix: http://www.w3.org/TR/hr-time/; type: typedef; text: DOMHighResTimeStamp
+url: http://www.w3.org/TR/CSS21/box.html#propdef-margin; type: attribute; text: margin
```
tabatkins: 'margin' is not an IDL attribute, it's a CSS property, and it'll autolink automatically if you use `'margin'`. (Single quotes trigger property/descriptor autolinking when the "css" set of shorthands is turned on, which they are by default.)

index.bs
```diff
@@ -8,14 +8,20 @@ Editor: Michael Blain, Google, mpb@google.com
...
+url: http://www.w3.org/TR/css3-values/#lengths; type: attribute; text: length
```
tabatkins: Similarly, "length" is not an IDL attribute. It look like you want the `<length>` production - autolink that with `<<length>>`.

index.bs
mpb: Is this enough?
Construct an *IntersectionObserverEntry*, passing in *time, rootBounds, boundingClientRect, intersectionRect* and *target*.

index.bs
```diff
@@ -8,14 +8,20 @@ Editor: Michael Blain, Google, mpb@google.com
...
+urlPrefix: http://www.w3.org/TR/html5/browsers.html; type: dfn; text: unit of related similar-origin browsing contexts
+url:http://w3c.github.io/requestidlecallback/#dfn-list-of-idle-request-callbacks; type: dfn; text: list of idle request callbacks
+url:http://heycam.github.io/webidl/#dfn-callback-this-value; type: dfn; text: callback this value
+urlPrefix:http://www.w3.org/TR/html51/webappapis.html; type: dfn; text: report the exception
+url:http://www.w3.org/TR/2014/WD-css-masking-1-20140213/#the-clip-path; type: dfn; text: clip path
```
tabatkins: You want to just use `'clip-path'` to autolink to the property.

index.bs
```diff
@@ -8,14 +8,20 @@ Editor: Michael Blain, Google, mpb@google.com
...
+url:https://dom.spec.whatwg.org/#concept-node-remove-ext; type: dfn; text: removing steps
+url:https://dom.spec.whatwg.org/#concept-tree-inclusive-descendant; type: dfn; text: inclusive descendants
+url:https://dom.spec.whatwg.org/#concept-tree-order; type: dfn; text: tree order
```
tabatkins: DOM is in the Bikeshed db. Did these really not work?

index.bs
mpb: Oh, magic!
'lengths' doesn't work the same way though.

index.bs
mpb: Done.

index.bs
tabatkins: There's no CSS property named "lengths", so `'lengths'` won't give you anything. ^_^

index.bs
mpb: Ah. removing steps, inclusive descendants and tree order actually do seem to work. I think callback this value and report the exception didn't.

mpb: Done.

index.bs
mpb: Done.

mpb:
Thanks for the reviews. Anything else outstanding, or does this look good to merge now?

domenic:
This looks great! Thanks very much for your attention to detail. This is a really solid spec now from my perspective. (Which, admittedly, is more focused on general spec integration than on the domain-specific issues.)

mpb:
Thanks!

PR #60, opened 2015-10-28
Update 'Threshold' to be a (double or sqeuence<double>)

mpb:
Update IntersectionObserverInit's 'threshold' property to be a (double or sequence<double>) instead of a single value.
This lets a single IntersectionObserver track multiple thresholds, and reduces the need for creating multiple IntersectionObservers to observe a single 'target'.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -147,14 +147,10 @@ The IntersectionObserver interface</h3>
                     <i>options</i>.{{root}}</li>
                 <li>Set <i>this</i>'s internal {{[[rootMargin]]}} slot to
                     <i>options</i>.{{rootMargin}}</li>
-                <li>Let <i>input</i> be a string equal to <i>options</i>.
+                <li>Let <i>thresholds</i> be a list equal to <i>options</i>.
                     {{threshold}}.</li>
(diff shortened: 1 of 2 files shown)


ojanvafai:
I talked this over with @tabatkins and @shans about this and they convinced me that we should just take numbers in the [0-1] range. So, 5% would be 0.05 instead and we can just use numbers without any need to do parsing. To allow multiple we can make it allow "<number> or sequence<number>".

mpb:
What's the language for that? Just add this to IntersectionObserverInit and drop the processing steps for parsing options.threshold?

**threshold, of type DOMString, defaulting to "0"**
Space-separated list of threshold(s) in the range of [0-1] at which to trigger callback. callback will be invoked when intersectionRect’s area changes from greater than or equal to any threshold to less than that threshold, and vice versa.

Threshold values are specified as numbers between 0 and 1 inclusive, and represent a percentage of the area as specified by target.getBoundingClientRect().

Note: "0" is effectively "any non-zero number of pixels".

ojanvafai:
lgtm, but the patch needs merging to tip of tree.

PR #71, opened 2015-11-06
Support multiple IntersectionObservers for a single target Element, add attributes and update algorithm.

### LINKED ISSUE #54: Queueing intersection observer tasks is pretty underdefined
It looks like the UA is allowed to post a task or to post an idle request callback.  Those have quite different observable behavior and it wouldn't surprise me if pages came to depend on one or the other in practice, leading them to work in some UAs but not others.

We should just pick a behavior here and define it, in my opinion.

### LINKED ISSUE #55: Don't call public API methods; instead call their backing algorithms
A few places in the spec refer to public API methods:

> The DOMRect corresponding to the target’s getBoundingClientRect().
> 
> Let boundingClientRect be the value of target.getBoundingClientRect(). 
> 
> For each observer in node’s internal [[RegisteredIntersectionObservers]] slot, invoke unobserve() on observer, passing in node as target. 

This is not quite the right thing to do, since it implies that e.g. if I do `Element.prototype.getBoundingClientRect = () => { throw new Error("boo"); }`, you'll call that overriden function.

Instead, the correct thing to do is to refer to the backing concept or algorithm behind that method. For unobserve, you'd probably define something in the processing model like

"To **remove an intersection observer** _o_ from an Element _target_, perform the following steps: ..."

Then unobserve() algorithm would do "Remove the intersection observer _this_ from _target_," and instead of doing "invoke unobserve()," you would do a similar thing.

For getBoundingClientRect(), things are more annoying, since CSSOM View hasn't factored themselves appropriately. There you could do something like "Let boundingClientRect be a DOMRectReadonly obtained by the same algorithm as that of Element's getBoundingClientRect() performed on target."

### LINKED ISSUE #57: Support multiple observers on a single target Element
The spec provides for a single PreviousIntersectionRect per target Element, which does not allow for multiple observers on a single target Element.

Proposal: RegisteredIntersectionObservers on the targetElement should be a sequence of:

{
    'observer': instance of IntersectionObserver,
    'previousVisibleRatio': ratio
}

### LINKED ISSUE #63: IntersectionObserver needs accessors for the IntersectionObserverInit values with which it was constructed
root()
rootMargin()
threshold()

mpb:
Fixes #54.
Fixes #55.
Fixes #57.
Fixes #63.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -171,15 +171,19 @@ The IntersectionObserver interface</h3>
                     {{[[ObservationTargets]]}} slot, return.</li>
                 <li>If |target| is NOT a descendent of |this|'s internal
                     {{[[root]]}} slot, <a>throw</a> a {{TypeError}}.</li>
-                <li>Add |this| to |target|'s internal
-                    {{[[RegisteredIntersectionObservers]]}} slot.</li>
+                <li>Let |intersectionObserverRegistration| be a dictionary with
+                    an <i>observer</i> property set to |this| and a
+                    <i>previousVisibleRatio</i> property set to <i>0</i>.</li>
+                <li>Append |intersectionObserverRegistration| to |target|'s
+                    internal {{[[RegisteredIntersectionObservers]]}} slot.</li>
                 <li>Add |target| to |this|'s internal {{[[ObservationTargets]]}}
                     slot.</li>
             </ol>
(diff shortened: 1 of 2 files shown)


mpb:
Ok, reverted "for each observer" change and made a note for rootBounds.
PTAL

Issue #81, opened 2016-01-20
Should IntersectionObserver => target be a bijection?


szager-chromium:
It has been suggested that the API would be more ergonomic if an IntersectionObserver represented an observation between a single root and a single target:

new IntersectionObserver(callback, targetElement, optionDict);

This would get rid of the observe() and unobserve() methods; disconnect() would remain.

Pros:

- More intuitive
- Requires less global bookkeeping

Cons:

- Less memory efficient (parameters like rootMargin are stored per-target, rather than per-observer).
- Callbacks would not be batched.  Currently, if an IntersectionObserver has multiple observed targets that generate notifications within a single cycle, the callback is invoked once with a sequence of notification records.  With the change, there would be one callback per target.

This is a brief summary of an email conversation; please add details.  The general sentiment is that the current spec is not a deal-breaker, but perhaps not ideal.  In practice, developers are free to use one observer per target, and shouldn't suffer any negative consequences.

An open question is whether, in practice, batching callbacks for multiple targets of a single observer is actually useful and more performant.  It's hard to know for certain until we see what developers do with it.

bmaurer:
My perspective as a site implementer -- the current spec makes it really unclear if it is important that a site batch using as few observers as possible. It's possible that different implementations will have different performance constraints.

Grouping elements by the parameters (such as rootMargin) requires an implementer to create a wrapper around the API. An API could either offer a fixed number of observers (here's one that has no margin, etc) which is brittle or have a cache of observers (requiring a fair amount of code). Alternatively the callers could make no effort to batch at all in which case InteractionObserver is just an overly verbose API.

If nothing else, it would be nice to have the spec clarify that browsers should expect to be able to support a high number of observers

szager-chromium:
You know, the difference between one-target-per-observer and many-targets-per-observer really comes down to the explicit calls to observe() and unobserve().  But for that, it's possible to simply ignore the difference.  So maybe one way to bridge the gap here is to allow an optional 'target' argument to the IntersectionObserver constructor, which would do an implicit observe(target); this:

var observer = new IntersectionObserver(callback, target, optionDict);

... would be equivalent to:

var observer = new IntersectionObserver(callback, optionDict);
observer.observe(target);

To create an observer without a first target, but with options, would be:

new IntersectionObserver(callback, null, optionDict);

[...]

PR #91, opened 2016-02-03
Some bikeshed cleanup, remove references to unrelated specs

mpb:
Right now, `<a>root<a/>` and `<a>target</a>` go to 
```
<a href="https://dom.spec.whatwg.org/#concept-tree-root">root</a>
<a href="https://dom.spec.whatwg.org/#concept-pi-target">target</a>
```
They should be local concepts within Intersection Observer.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -38,7 +38,6 @@ url: http://dev.w3.org/csswg/cssom-view/#pinch-zoom; type:dfn; text: pinch zoom
 <pre class="link-defaults">
 spec:dom-ls
 	type:interface; text:Document
-	type:dfn; for:tree; text:root
 spec: css-values-3; type: dfn
 	text: absolute length
 	text: dimension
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -131,28 +131,25 @@ The IntersectionObserverCallback</h3>
...
-{{IntersectionObserverInit/rootMargin}}
-or {{threshold}} values
-for each <a>target</a>
-seems to introduce more complexity without solving additional use-cases.
-Per-{{observe()}} options could be provided in the future if V2 introduces a need for it.
+The {{IntersectionObserver}} interface can be used to observe changes in the
+intersection of a <a>target</a> {{Element}} and a {{IntersectionObserver/root}}
+{{Element}} (or the top-level document's viewport).
+
+Note: In {{MutationObserver}}, the {{MutationObserverInit}} options are passed
```
szager-chromium: This section should be updated based on the comments in issue #81, but that can go in a different patch.

mpb: Agreed, but there still doesn't seem to be consensus on that thread?

szager-chromium: I think the consensus is:

For typical usage, developers should feel free to create one observer per target.

In the rare case where many targets are observing the same root with the same parameters; and there is a likelihood that multiple notifications will be generated during a single run of the algorithm; then the grouping of notifications in the callback may offer a performance advantage.

mpb: Ok, but none of that changes the API as spec'ed, right?

szager-chromium: Correct; it's just to anticipate a likely faq and source of developer confusion.

Issue #74, opened 2015-11-21
Define the document scope of an observer


szager-chromium:
Proposal from conversation with ojan:

If an observer is created with a root, targets are limited to the hierarchy below the root.

If an observer is created without a root, targets are limited to the hierarchy below the main frame of the execution context in which the observer was created.

mpb:
SGTM

tabatkins:
Throw a HierarchyError, I guess?

PR #92, opened 2016-02-04
Take a stab at specifying observer scope.

szager-chromium:
Issue #74.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5,7 +5,7 @@ ED: https://github.com/WICG/IntersectionObserver/
 Shortname: intersection-observer
 Level: 1
 Editor: Michael Blain, Google, mpb@google.com
-Abstract: This specification describes  an API that can be used to understand the visibility and position of DOM elements relative to an ancestor container or to the top-level window's visual viewport. The position is delivered asynchronously and is useful for understanding the visibility of elements and implementing pre-loading and deferred loading of DOM content.
+Abstract: This specification describes  an API that can be used to understand the visibility and position of DOM elements relative to an ancestor container or to the top-level window's viewport. The position is delivered asynchronously and is useful for understanding the visibility of elements and implementing pre-loading and deferred loading of DOM content.
 Group: Web Performance Working Group
 Repository: WICG/IntersectionObserver
 </pre>
@@ -40,7 +40,13 @@ urlPrefix: https://heycam.github.io/webidl/
 	urlPrefix: #dfn-; type:dfn; text: throw
 	urlPrefix: #idl-; type:interface; text: double
 url: https://drafts.csswg.org/css2/box.html#content-edge; type:dfn; text: content-box
-url: http://dev.w3.org/csswg/cssom-view/#pinch-zoom; type:dfn; text: pinch zoom
+urlPrefix: https://drafts.csswg.org/css-box/
+	url: #viewport; type: dfn; text: viewport
+	url: #containing-block; type: dfn; text: containing block
+	url: #content-area; type: dfn; text: content area
+urlPrefix: http://dev.w3.org/csswg/cssom-view/
+	url: #pinch-zoom; type:dfn; text: pinch zoom
+	url: #dom-element-getboundingclientrect; type:dfn; text: getBoundingClientRect
 </pre>
 
 <pre class="link-defaults">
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -138,8 +147,29 @@ intersection with the <a>intersection root</a>, as per the
 The IntersectionObserver interface</h3>
 
 The {{IntersectionObserver}} interface can be used to observe changes in the
-intersection of a <a>target</a> {{Element}} and a {{IntersectionObserver/root}}
-{{Element}} (or the top-level document's viewport).
+intersection of a {{IntersectionObserver/root}} {{Element}} (or the
+<a>top-level browsing context</a>'s visual viewport), and one or more <a>target</a> {{Elements}}.
```
mpb: Elements point to https://dom.spec.whatwg.org/#elements which is an array-like collection.

szager-chromium: Changed this to {{Element}}s, here and elsewhere

mpb:
LGTM

szager-chromium:
Ojan, any thoughts on this?

index.bs
```diff
@@ -138,8 +147,29 @@ intersection with the <a>intersection root</a>, as per the
...
 The {{IntersectionObserver}} interface can be used to observe changes in the
-intersection of a <a>target</a> {{Element}} and a {{IntersectionObserver/root}}
-{{Element}} (or the top-level document's viewport).
+intersection of a {{IntersectionObserver/root}} {{Element}} (or the
+<a>top-level browsing context</a>'s visual viewport), and one or more <a>target</a> {{Element}}s.
+
+An {{IntersectionObserver}} with a {{IntersectionObserver/root}} {{Element}} can
+observe any <a>target</a> {{Element}} that is a descendant of the
+{{IntersectionObserver/root}} in the <a>node tree</a>, including {{Element}}s inside
+an <a>iframe</a> that belongs to the same <a>unit of related similar-origin browsing contexts</a>
```
ojanvafai: As per recent issue discussions, this text isn't right anymore, right? Namely:
1. Elements inside similar-origin browsing context frames are no longer allowed.
2. Root needs to be an ancestor of target in the containingBlock chain.

index.bs
```diff
@@ -138,8 +147,29 @@ intersection with the <a>intersection root</a>, as per the
...
+     <pre highlight="html">
+     &lt;div id="root"&gt;
+       &lt;iframe srcdoc="&lt;div id='target'&gt;"&gt;&lt;iframe&gt;
+     &lt;div&gt;
+     </pre>
+</div>
+
+An {{IntersectionObserver}} with no {{IntersectionObserver/root}} {{Element}}
+will automatically observe intersections with the <a>top-level browsing context</a>'s
+visual viewport, and valid <a>targets</a> include any {{Element}} in the
```
ojanvafai: I'm not sold that we should be changing from layout to visual viewport. I think we should probably VC with me, you and @esprehn to hash it out.

index.bs
szager-chromium: Fixed.

szager-chromium:
Added a definition for "implicit root", and used that term throughout.

PTAL

index.bs
```diff
@@ -70,32 +87,27 @@ however, including (but not limited to):
...
+browser architectures in the face of filters, webgl, and other features). In all
+of these scenarios the information is useful even when delivered at a slight
+delay and without perfect compositing-result data.
+
+The Intersersection Observer API addresses the above issues by giving developers
+a new method to asynchronously query the position of an element with respect to
+other elements or the global viewport. The asynchronous delivery eliminates the
+need for costly DOM and style queries, continuous polling, and use of custom
+plugins. By removing the need for these methods it allows applications to
+significantly reduce their CPU, GPU and energy costs.
```
tabatkins: I can't tell what you did or didn't change here.  When I generally reformatted the spec I switched to [semantic linefeeds](http://rhodesmill.org/brandon/2012/one-sentence-per-line/), which I use in all of my specs and which make diffs better.

If you didn't actually change anything, can you revert this back to the formatting it had?

index.bs
```diff
@@ -18,21 +20,35 @@ url: http://w3c.github.io/requestidlecallback/#dfn-list-of-idle-request-callback
...
+		type: dfn; text: nested browsing contexts
 	urlPrefix: webappapis.html;
 		type: dfn; text: report the exception
 		type: dfn; text: event loop
 	urlPrefix: infrastructure.html;
 		type: dfn; text: rules for parsing dimension values
 		url: #dfn-callback-this-value; type: dfn; text: callback this value
+	urlPrefix: embedded-content.html;
+		url: #the-iframe-element; type: dfn; text: iframe
 urlPrefix: https://heycam.github.io/webidl/
```
tabatkins: You can remove this - Bikeshed already knows about the iframe anchor.  (Also, type:dfn is wrong - it's type:element.)

index.bs
```diff
@@ -41,6 +57,7 @@ spec:dom-ls
 spec: css-values-3; type: dfn
 	text: absolute length
 	text: dimension
+spec: embedded-content; type: element; text: iframe
```
tabatkins: There's no spec called "embedded-content".  Bikeshed only knows of one "iframe" reference, so you shouldn't need this line at all.  (If you didn't know, there's a markup shorthand for element refs - `<{iframe}>`.)

tabatkins:
The formatting no longer matches the rest of the spec, but otherwise it looks reasonable.

ojanvafai:
lgtm

index.bs
szager-chromium: Done

index.bs
szager-chromium: Done

Issue #70, opened 2015-11-06
Clarify how clipping is computed when rootMargin is set


szager-chromium:
What happens when a target element transitions from outside of rootMargin to within rootMargin of the root or visual viewport, but it is clipped by one of its other ancestors?

szager-chromium:
ojan@ says: if the target is clipped by something other than the root element (or viewport), then it should *not* generate intersection events.

ojanvafai:
Lets add a note to the processing model to this effect to make it clear.

PR #93, opened 2016-02-04
Clarify that rootMargin is only applied to root.

szager-chromium:
Issue #70

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -174,11 +174,9 @@ interface IntersectionObserver {
 		3. Set |this|.{{IntersectionObserver/root}} to
 			|options|.{{IntersectionObserverInit/root}}
 		4. Attempt to <a>parse a root margin</a>
-			from |options|.{{IntersectionObserverInit/rootMargin}}.
-			If a list is returned,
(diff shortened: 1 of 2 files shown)


mpb:
|options.rootMargin| doesn't look correct in the bikeshed output.

index.bs
```diff
@@ -174,11 +174,9 @@ interface IntersectionObserver {
 		3. Set |this|.{{IntersectionObserver/root}} to
 			|options|.{{IntersectionObserverInit/root}}
 		4. Attempt to <a>parse a root margin</a>
-			from |options|.{{IntersectionObserverInit/rootMargin}}.
-			If a list is returned,
+			from |options.rootMargin|.  If a list is returned,
```
mpb: This line doesn't seem to translate correctly into _options.rootMargin_ in the output html.

szager-chromium: Fixed, also added some more verbiage to the threshold attribute description.

szager-chromium:
ping

ojanvafai:
lgtm

Issue #84, opened 2016-02-01
should ancestor chain requirement for active observations be containingBlock or DOM-based?


ojanvafai:

esprehn:
At the very least this should use the Shadow DOM composed tree I think?

ojanvafai:
The tricky case here is something like the following:
<div style="position:absolute">
  <div id="root">
    <div id="target" style="position:absolute; top:0; left:0;">
    </div>
  </div>
</div>

Should you be able to observe target even though it's positioned WRT an ancestor of the root?

ojanvafai:
I think the above case ideally would not be allowed, but the case below is less clear to me since the target is statically positioned.

'''
<div style="position:absolute">
  <div id="root">
    <div id="target" style="position:absolute;">
    </div>
  </div>
</div>
'''

szager-chromium:
All signs point to containing block.

ojanvafai:
Turns out the target in the latter case doesn't scroll when you scroll the root. So, yeah, containingBlock SGTM.

Issue #52, opened 2015-10-09
rootBounds leaks information in the cross-origin case


bzbarsky:
Does it allow percentage values?  `calc()`?  Units other than `px`?  If if allows `em` and `ex` units, which font size is used for them, and when?  If it allows `vh` and `vw` units, which viewport is used for them and when?

mpb:
Is there standard terminology for this anywhere, or it's always ad-hoc?

bzbarsky:
Standard terminology in what sense?

The CSS specs always define, for each property that can be a length, what the various relative units are relative to.  CSS doesn't need to worry about the "when" because it just assumes that the world is all updated at once, but this spec is snapshotting things at particular times, so depends on timing.

ojanvafai:
WRT what to resolve em/ex/vh/vw to, my gut feeling is that they should resolve relative to the "root" element (as per the root constructor argument) and if root is undefined, then the initial containing block. I guess the thing I hadn't considered here is that this means that style changes on the root element force you to recompute the margin. I think that's OK?

In terms of timing, I think we should expect these to force style to be up to date before computing these values. The current snapshot time (i.e. end of the "update the rendering" step of the HTML event loop processing model) should already have CSS all be up to date.

bzbarsky:
> then the initial containing block

The root defaults to the topmost viewport, no?  That's not information you can normally get cross-origin, but it seems like if we allow computing vh/vw relative to it, then you can read the size of the toplevel viewport from a cross-origin iframe.

Are there even sane use cases for allowing vh/vw here?

ojanvafai:
The size of the top level viewport is already largely exposed in every browser via other channels (e.g. requestAnimationFrame throttling of offscreen content). This gives a bit more fidelity, but I'm not sure it really leaks anything meaningfully more.

That said, I'm supportive of limiting the types we allow here. It's em/ex that I think really don't have use-cases actually.

vh/vw can make sense to me if you're thinking about lazy-loading/infinite-scroll use-cases since scroll velocity is proportional to the size of the window. I think?

bzbarsky:
> e.g. requestAnimationFrame throttling of offscreen content

Which is (1) optional and (2) doesn't actually expose that as far as I can tell.  Especially not if your iframe happens to be in another scroll container.

If you care about scroll use cases, then you care about vw/vh relative to the thing that's scrolling, which may or may not be the root as this spec defines it.

tabatkins:
```
[...]

PR #97, opened 2016-02-14
Containing block

szager-chromium:
Issue #84: root must be containing block ancestor of target.
Issue #81: OK to create one observer per target.
Issue #52: percent rootMargins not allowed in the cross-origin case.


szager-chromium:
(for the record, lgtm'ed by slightlyoff)

slightlyoff:
Clarified in discussion that the effect of this is only to prevent generation/delivery of records in the case when  roots aren't containing blocks (although perhaps we should add a non-normative note suggesting that devtools warn in these cases). LGTM++

Issue #62, opened 2015-10-30
properly spec clip rect


ojanvafai:
"ISSUE2 TBD: Do the clip rects include clip-path and other clipping properties or just overflow clipping?"

@tabatkins and I just discussed this and we think that we should only take overflow clipping into account. That's the only reasonable think for the non-malicous use cases. For the malicious ones we're going to need to do something else entirely anyways since it need to take masks, etc into account (any masked region would count as clipped regardless of what the mask does).

So, the only thing we need here is to remove ISSUE 2 and to find a well-defined definition of "clip rect" for overflow clipping. @tabatkins says he can help with the latter.

szager-chromium:
Upon further discussion, let's honor overflow and css clipping, but not masking or occlusion.

PR #98, opened 2016-02-14
Add algorithm for computing intersection rect.

szager-chromium:
Issue #62 

Not sure if this level of detail is too agonizing or not, please advise.


slightlyoff:
LGTM

Issue #87, opened 2016-02-01
webkit would prefer that we not allow cross-frame root-->target observations except for the case of the null root


ojanvafai:

szager-chromium:
Any particular reason for this?

KenjiBaheux:
I'm worried this will mean that other third parties in a chain will have to keep their legacy Viewable Impression measurement code in place (i.e. Advertisers and extra Ads exchange want to measure Viewability on their own).

ojanvafai:
They don't think the implementation complexity is justified since we have no use cases for it.

@KenjiBaheux, that case should be fine still. Same-origin, cross-frame with a null root works. It's just that you can't have a specified root in a different frame from the target being observed. Which is to say, if they're in different frames, they're handled the same as if the target is not a descendant of the root.

szager-chromium:
There's no known use case for this, so let's go ahead and do it.

Issue #72, opened 2015-11-07
Need to define actual syntax for rootMargin


mpb:
Does it allow percentage values? `calc()`? Units other than `px`? If if allows `em` and `ex` units, which font size is used for them, and when? If it allows `vh` and `vw` units, which viewport is used for them and when?

cloned from #52 which went off-topic.
cc: @bzbarsky

mpb:
Maybe can be % or `px`, and if you don't specify a `root` then % should throw.

szager-chromium:
Consensus in favor of mpb's suggestion, let it be so.

ojanvafai:
Actually, on further thought, I think we should just disallow rootMargin if root is null. We don't have clear use-cases for it or authors asking for it. The reasons for this are:
1. Makes it consistent with having rootBounds be null if root is null.
2. For browsers that don't run the rendering pipeline for off-screen, cross-origin frames, this avoids needing to run the rendering pipeline just to see if something observed inside the frame crosses the threshold.

Any objections? @tabatkins @esprehn @szager-chromium @mpb

mpb:
No objections.

esprehn:
We discussed this in person with Apple, and I don't think that's correct, it means you can't specify a runway for top level observers in the common case, you end up having to observe the documentElement, but the documentElement can actually be 0x0 if the page is all positioned so that's not right either.

This needs something more subtle like "rootMargin only applies in the top level frame, or when a root is specified in a subframe."

ojanvafai:
Forgot to comment on this one. I commented on https://github.com/WICG/IntersectionObserver/issues/52#issuecomment-177743457.

So we'll have two things:
1. rootMargins/rootBounds don't work for cross-origin observations as per issue #52.
2. same-origin, cross-frame observation doesn't work as per issue #87.

That only leave same-frame observations, which allows px and % values in rootMargin.

@esprehn that sound OK?

szager-chromium:
Wait, that's missing something:

3. same-origin, cross-frame observation with null root.  Does rootMargin work for this?

ojanvafai:
@szager-chromium yes, I believe that's the case @esprehn was trying to point out as that's something you don't have access to from JS, so the only way to observer the magical root is to use the null root.

szager-chromium:
@ojanvafai I am confused.  According to #87, any cross-frame observation must have a null root.  There are two possibilities for cross-frame observations:

1. cross-origin -- no rootBounds, no rootMargin
[...]

Issue #58, opened 2015-10-21
Remove auto-unobserving. Instead have active and inactive observations.


szager-chromium:
Proposal: IntersectionObserver.observe should take a second, optional argument (let's call it unobserveCallback) which is a callback to invoke whenever the 'unobserve' steps are run for this observation.  Section 3.2.5 can be removed from the spec (i.e., nothing is done during node removal).  When the intersection observation steps are run (section 3.2.4), a check should be made that the target is still a descendant of the root.  If the check fails, the 'unobserve' steps are run, invoking unobserveCallback.

This proposal provides an additional benefit for the case where the target and/or root elements are moved in the DOM tree, but their ancestor/descendant relationship stays intact (e.g., sorting siblings).  In this case, under the current spec, any observer relationships between target and root would be destroyed without notification.  With this proposal, the observer relationships would stay intact.

ojanvafai:
@tabatkins and @shans and I discussed this and we think it's too weird to have the unobserve happen when the intersection observation steps are run. For example, what if we decide to run those steps on the compositor thread in the future?

Unobserving immediately upon removal (as we currently have specced) along with the unobserve callback should be ergonomic enough and doesn't lead to a bunch if implementation complexity/requirements.

Sound good?

@esprehn @mpb @slightlyoff

szager-chromium:
I'll defer to the voice of experience here, but I would point out two things:

- As written, when a node is removed, we have to do a full tree walk of its descendants to look for observer targets.  That seems like too much work.

- Intuitively, I would expect to be able to remove and immediately reinsert an observer root (or one of its ancestors) without invalidating all of the root's intersection observers, since I didn't alter the ancestry relationship between target and root.  Adding an unobserveCallback makes this a bit more palatable, but it still feels fragile.

szager-chromium:
I updated the title of this issue to focus on adding an unobserveCallback.  The question of whether to keep observations intact if a root is removed and reinserted without breaking the relationship between root and target can be considered separately.

unobserveCallback is needed so that implicit unobserve's caused by DOM mutation won't go unreported.  Yet to be determined: the timing of the callback; it need not be synchronous.

mpb:
I think calling an explicit unobserve() doesn't need to trigger the unobserved callback, right? Is there precedent for this in other APIs?

ojanvafai:
[...]

PR #100, opened 2016-02-16
Clarify rootMargin processing and active state.

szager-chromium:
- rootMargin only honored for same-origin.
- If containing block or target-in-same-frame-as-explicit-root
  requirements are violated, don't generate notifications.
- Get rid of per-browsing-context-unit IntersectionObservers list.
- Update text about HTML Processing Model.

Issue #58
Issue #87
Issue #72

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -218,9 +218,7 @@ interface IntersectionObserver {
 		7. Sort |thresholds| in ascending order.
 		8. If |thresholds| is empty, append <code>0</code> to |thresholds|.
 		9. Set |this|.|thresholds| to |thresholds|.
-		10. Append |this| to the <a>unit of related similar-origin browsing context</a>'s
-			<a>IntersectionObservers</a> list.
-		11. Return |this|.
+		10. Return |this|.
 	: <dfn>observe(target)</dfn>
 	::
 		1. If |target| is in |this|'s internal {{[[ObservationTargets]]}} slot,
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -582,49 +583,45 @@ To <dfn>run the update intersection observations steps</dfn> for an
...
 1. Let |unit| be the <a>unit of related similar-origin browsing contexts</a> for
 	|loop|.
-2. For each |observer| in |unit|'s <a>IntersectionObservers</a> list:
+2. Let |observer list| be a list of all {{IntersectionObserver}}s
+	whose {{IntersectionObserver/root}} is in the DOM tree of a document in |unit|.
+3. For each |observer| in |observer list|:
 	1. Let |rootBounds| be |observer|'s <a>root intersection rectangle</a>.
 	2. For each |target| in |observer|'s internal {{[[ObservationTargets]]}} slot:
-		1. Let |targetRect| be a {{DOMRectReadOnly}}
+		1. If |target| is not a descendant of the <a>intersection root</a>
```
ojanvafai: This should also only apply if the <a>intersection root</a> is not the <a>implicit root</a>.

ojanvafai:
lgtm once that one issue is addressed.

index.bs
szager-chromium: Done.

PR #106, opened 2016-02-25
Use a document-centric processing model.

szager-chromium:
This should integrate better with the existing html event loop spec.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8,7 +8,7 @@ Editor: Michael Blain, Google, mpb@google.com
 Abstract: This specification describes  an API that can be used to understand the visibility and position of DOM elements ("targets") relative to a containing element or to the top-level viewport ("root"). The position is delivered asynchronously and is useful for understanding the visibility of elements and implementing pre-loading and deferred loading of DOM content.
 Group: Web Performance Working Group
 Repository: WICG/IntersectionObserver
-Ignored Vars: rootMargin
+Ignored Vars: rootMargin, docs
 </pre>
 
 <pre class="anchors">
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -468,12 +472,12 @@ the <a>Intersection Observer</a> API.
...
-<h4 id='browsing-contexts-defines'>
-Browsing Contexts</h4>
+<h4 id='document-defines'>
+Document</h4>
 
-Each <a>unit of related similar-origin browsing contexts</a> has an
-<dfn for="browsing context">IntersectionObserverTaskQueued</dfn> flag which
-is initialized to false.
+Each <a>Document</a> has an
+<dfn for="browsing context">IntersectionObserverTaskQueued</dfn> flag
```
mpb: Is \<dfn for="browsing context"> still accurate?

index.bs
```diff
@@ -543,15 +547,16 @@ similar-origin browsing contexts</a> |unit|, run these steps:
 <h4 id='queue-intersection-observer-entry-algo'>
 Queue an IntersectionObserverEntry</h4>
 
-To <dfn>queue an IntersectionObserverEntry</dfn> for |observer|, given a
-<a>unit of related similar-origin browsing contexts</a> |unit|, |observer|,
-|time|, |rootBounds|, |boundingClientRect|, |intersectionRect| and |target|,
+To <dfn>queue an IntersectionObserverEntry</dfn> for an {{IntersectionObserver}} |observer|, given a
+<a>Document</a> |document|; {{DOMHighResTimeStamp}} |time|;
+{{DOMRect}}'s |rootBounds|, |boundingClientRect|, and |intersectionRect|;
```
mpb: Why is DOMRect's possessive?

mpb:
Looks good to me, modulo some tiny style nits.

index.bs
szager-chromium: Changed to <dfn for="Document">

index.bs
szager-chromium: Removed the apostrophe.  English is hard.

domenic:
Interesting. This changes not only the "run the update intersection observations steps" to be document-centric, but the whole spec. This now mismatches mutation observers---but maybe that's OK?

szager-chromium:
@domenic I didn't refer to the MutationObserver spec when making this change, but I think the whole thing reads better when it only refers to documents.  WDYT?

domenic:
It's hard for me to say. Maybe @ajklein or @annevk know why mutation observer is in terms of event loops/units of related similar-origin browsing contexts, instead of in terms of documents?

It seems quite possible that the mutation observer design is not appropriate for intersection observers, and documents is better, but it's worth checking first.

ajklein:
@annevk probably knows more precisely, but MutationObserver, implementation-wise, is in terms of "changes that have occurred on this turn on this thread", which happens to map to "similar-origin browsing contexts".

annevk:
@domenic mutation observers is this way because mutations can involve multiple documents (e.g., when you append a node that is from another document).

PR #107, opened 2016-02-25
Don't specify behavior of observer with invalid root.

szager-chromium:
The concept of a node being "deleted", as opposed to merely being
detached from the DOM, is not specified by any standard, so we
shouldn't refer to it here.


ojanvafai:
lgtm

PR #116, opened 2016-03-16
Add IntersectionObserver polyfill

surma:
r: @ojanvafai 
cc: @ianvollick @paullewis @flackr

This adds a polyfill for the IntersectionObserver API. It is compatible to the current version of the spec but ignores margins for now, as I wanted to avoid implementing a CSS string parser.

There’s also a simple test suite by @ojanvafai. All tests pass except for the margin tests for the reasons mentioned above. Please take a look at the tests to check that the polyfill is behaving as you’d expect.
screenshot 2016-03-16 14 05 35

Any other feedback obviously welcome as well :)

### DIFF
--- a/polyfill/intersectionobserver-polyfill.js
+++ b/polyfill/intersectionobserver-polyfill.js
@@ -112,7 +112,7 @@ limitations under the License.
     _update: function() {
       var rootRect = this._rootRect();
       this._observationTargets.forEach(function(oldIntersectionEntry, target) {
-        var targetRect = target.getBoundingClientRect();
+        var targetRect = augmentRect(target.getBoundingClientRect());
         var intersectionRect = intersectRects(rootRect, targetRect);
         if(!intersectionRect) {
           return;
@@ -159,7 +159,7 @@ limitations under the License.
 
     _rootRect: function() {
       if(this._root) {
-        return this.root.getBoundingClientRect();
+        return augmentRect(this.root.getBoundingClientRect());
       }
       return {
         top: 0,
@@ -221,5 +221,11 @@ limitations under the License.
         fn.apply(context, args);
       }, delay);
     };
+  };
+
+  var augmentRect = function(r) {
+    r.width = r.width || r.right - r.left;
+    r.height = r.height || r.bottom - r.top;
+    return r;
   };
 })();
--- a/polyfill/intersectionobserver-polyfill.js
+++ b/polyfill/intersectionobserver-polyfill.js
@@ -1,0 +1,188 @@
+(function() {
+  // TODO: rootMargin are completely ignored for now
+  
+  // Constructor
+  window.IntersectionObserver = function(callback, options) {
+    options = options || {};
+    
+    if(!(callback instanceof Function)) {
+      throw('callback needs to be a function');
+    }
+    
+    this._callback = callback;
+    this._root = options.root || null;
+    this._rootMargin = options.rootMargin || [0, 0, 0, 0];
+    this._thresholds = options.threshold || 0;
+    this._init();
+  };
+
+  window.IntersectionObserver.prototype = {
+    //
+    // Public API
+    //
+    get root() {
+      return this._root || document;
+    },
+    
+    get rootMargin() {
+      return '0';
+    },
+    
+    get thresholds() {
+      // 0 means call callback on every change
+      // See note at http://rawgit.com/WICG/IntersectionObserver/master/index.html#intersection-observer-init
+      if(this._thresholds === 0) {
+        return 0;
+      }
+      if(this._thresholds instanceof Array) {
+        return this._thresholds;
+      }
+      return [this._thresholds];
+    },
+    
+    observe: function(target) {
+      if(!(target instanceof HTMLElement)) {
+        throw('Target needs to be an HTMLelement');
+      }
+      
+      var root = this.root;
+      var ancestor = target.parentNode;
+      while (ancestor != root) {
+        if (!ancestor) {
+          throw('Target must be decendant of root');
+        }
+        ancestor = ancestor.parentNode;   
+      }
+      
+      this._observationTargets.set(target, {});
+    },
+    
+    unobserve: function(target) {
+      this._observationTargets.delete(target);
+    },
+    
+    disconnect: function() {
+      this._observationTargets.clear();
+      this.root.removeEventListener('scroll', this._boundUpdate);
+      window.clearInterval(this._timeoutId);
+      this._descheduleCallback();
+    },
+    
+    takeRecords: function() {
+      this._update();
+      this._descheduleCallback();
+      var copy = this._queue.slice();
+      this._queue = [];
+      return copy;
+    },
+    
+    //
+    // Private API
+    //
+    _init: function() {
+      this._observationTargets = new Map();
+      this._boundUpdate = this._update.bind(this);
+      this.root.addEventListener('scroll', this._boundUpdate);
+      this._intervalId = window.setInterval(this._boundUpdate, 100);
+      this._queue = [];
+    },
+    
+    _update: function() {
+      var rootRect = this._rootRect();
+      this._observationTargets.forEach(function(oldIntersectionEntry, target) {
+        var targetRect = target.getBoundingClientRect();
+        var intersectionRect = intersectRects(rootRect, targetRect);
+        if(!intersectionRect) {
+          return;
+        }
+        var targetArea = targetRect.width * targetRect.height;
+        var intersectionArea = intersectionRect.width * intersectionRect.height;
+        var intersectionRatio = intersectionArea / targetArea;
+        if(!this._hasCrossedThreshold(oldIntersectionEntry.intersectionRatio, intersectionRatio)) {
+          return;
+        }
+        var intersectionEntry = {
+          time: window.performance.now(),
+          rootBounds: rootRect,
+          boundingClientRect: targetRect,
+          intersectionRect: intersectionRect,
+          intersectionRatio: intersectionRatio,
+          target: target
+        };
+        Object.freeze(intersectionEntry);
+        this._queue.push(intersectionEntry);
+        this._scheduleCallback();
+        this._observationTargets.set(target, intersectionEntry);
+      }.bind(this));
+    },
+    
+    _scheduleCallback: function() {
+      if(this._idleCallbackId) {
+        return;
+      }
+      this._idleCallbackId = window.requestIdleCallback(function() {
+        this._descheduleCallback();
+        this._callback(this._queue, this);
+        this._queue = [];
+      }.bind(this), {timeout: 100});
+    },
+    
+    _descheduleCallback: function() {
+      if(!this._idleCallbackId) {
+        return;
+      }
+      window.cancelIdleCallback(this._idleCallbackId);
+      this._idleCallbackId = null;
+    },
+    
+    _rootRect: function() {
+      if(this._root) {
+        return this.root.getBoundingClientRect();
+      }
+      return {
+        top: 0,
+        left: 0,
+        right: window.innerWidth,
+        width: window.innerWidth,
+        bottom: window.innerHeight,
+        height: window.innerHeight
+      };
+    },
+    
+    // FIXME: so hack, much performance, very JSON
+    _hasCrossedThreshold: function(oldRatio, newRatio) {
+      if(this.thresholds === 0) {
+        return newRatio != oldRatio;
+      }
+      var b1 = this.thresholds.map(function(threshold) {
+        return threshold <= oldRatio;
+      });
+      var b2 = this.thresholds.map(function(threshold) {
+        return threshold <= newRatio;
+      });
+      return JSON.stringify(b1) !== JSON.stringify(b2);
+    }
+  };
+
+  var intersectRects = function(rect1, rect2) {
+    var top = Math.max(rect1.top, rect2.top);
+    var bottom = Math.min(rect1.bottom, rect2.bottom);
+    var left = Math.max(rect1.left, rect2.left);
+    var right = Math.min(rect1.right, rect2.right);
+    var intersectionRect = {
+      top: top,
+      bottom: bottom,
+      left: left,
+      right: right,
+      width: right-left,
+      height: bottom-top
+    };
+    if(top > bottom) {
+      intersectionRect.height = 0;
+    }
+    if(left > right) {
+      intersectionRect.width = 0;
+    }
+    return intersectionRect;
+  };
+})();
--- a/polyfill/intersectionobserver-polyfill.js
+++ b/polyfill/intersectionobserver-polyfill.js
@@ -53,7 +53,7 @@
       var ancestor = target.parentNode;
       while (ancestor != root) {
         if (!ancestor) {
-          throw('Target must be decendant of root');
+          throw('Target must be descendant of root');
         }
         ancestor = ancestor.parentNode;
       }
--- a/polyfill/intersectionobserver-polyfill.js
+++ b/polyfill/intersectionobserver-polyfill.js
@@ -1,3 +1,15 @@
+/*
+Copyright 2016 Google Inc. All Rights Reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+   http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
 (function() {
   // TODO: rootMargin are completely ignored for now
 
@@ -197,17 +209,17 @@
   };
 
   var debounce = function(fn, delay) {
-  var timer = null;
-  return function () {
-    if(timer) {
-      return;
-    }
-    var context = this, args = arguments;
-    clearTimeout(timer);
-    timer = setTimeout(function () {
-      timer = null;
-      fn.apply(context, args);
-    }, delay);
+    var timer = null;
+    return function () {
+      if(timer) {
+        return;
+      }
+      var context = this, args = arguments;
+      clearTimeout(timer);
+      timer = setTimeout(function () {
+        timer = null;
+        fn.apply(context, args);
+      }, delay);
+    };
   };
-}
 })();
--- a/polyfill/intersectionobserver-polyfill.js
+++ b/polyfill/intersectionobserver-polyfill.js
@@ -86,7 +86,7 @@
     _init: function() {
       this._observationTargets = new Map();
       this._boundUpdate = this._update.bind(this);
-      this.root.addEventListener('scroll', this._boundUpdate);
+      this.root.addEventListener('scroll', debounce(this._boundUpdate, 100));
       this._intervalId = window.setInterval(this._boundUpdate, 100);
       this._queue = [];
     },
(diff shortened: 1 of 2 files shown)


polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,188 @@
...
+  // Constructor
+  window.IntersectionObserver = function(callback, options) {
+    options = options || {};
+    
+    if(!(callback instanceof Function)) {
+      throw('callback needs to be a function');
+    }
+    
+    this._callback = callback;
+    this._root = options.root || null;
```
mpb: If 'root' is passed in, there should probably be some check that it's valid, and not an object or function or something random.

surma: All `observe()` calls will fail if `root` is invalid. So this check is not strictly needed, but I agree it’s probably a good idea to add that check.

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,188 @@
...
+    observe: function(target) {
+      if(!(target instanceof HTMLElement)) {
+        throw('Target needs to be an HTMLelement');
+      }
+      
+      var root = this.root;
+      var ancestor = target.parentNode;
+      while (ancestor != root) {
+        if (!ancestor) {
+          throw('Target must be decendant of root');
```
mpb: s/decendant/descendant

mpb:
Should this polyfill throw an exception or otherwise indicate an error if someone tries to use it in a cross-domain iframe with `root` === `null`? In all other situations, I think this polyfill will provide correct behavior, but for this case the behavior will diverge pretty substantially.

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,188 @@
...
+    },
+    
+    //
+    // Private API
+    //
+    _init: function() {
+      this._observationTargets = new Map();
+      this._boundUpdate = this._update.bind(this);
+      this.root.addEventListener('scroll', this._boundUpdate);
+      this._intervalId = window.setInterval(this._boundUpdate, 100);
```
triblondon: This seems less efficient than typical lazyload libraries, in that it polls the state of the observed element on every scroll event, and every 100ms. I would like to put this polyfill into the polyfill.io service library but am concerned about the frequency of these events. How about throttling the callback to one invocation per 100ms?

Also I assume we're polling separately to the scroll events to capture scenarios where the element moves from outside to inside the viewport without the document scroll position changing. Could this lean on MutationObserver to avoid setting up a tight interval timer that runs indefinitely?

mpb: That's a good point, "every scroll event" should probably be throttled to 100ms too.

surma: So, I agree that the polling and the scroll-events overlap. I can do something about that.

However, I am not sure about rate-limiting the updates to 100ms in general. If you want to use this data in animation, you ideally get an update from the IntersectionObserver every frame, right? Or did I miss something in the spec about this? I was thinking to use `requestAnimationFrame()` so you get frame-by-frame updates in the polyfill.

mpb: Hm it would be great if the library can collect events as often as necessary, but batch delivering them and cap invoking the callback to no more than once every 100ms. Unfortunately, calling getBoundingClientRect on every scroll event is expensive. Maybe someone with more knowledge about the performance implications can weigh in? @KenjiBaheux @ojanvafai ?

surma: `getBoundingClientRect()` is expensive, but that’s the polyfill only. The actual browser implementation will be more efficient and will probably dispatch on every frame, right? Otherwise the entire thing becomes rather useless for 60fps animations. Also, I thought you would use the thresholds to limit the number of events.

It’s a tough call and I’d appreciate more feedback. Personally, I’d rather emulate the same behavior of an actual browser implementation than worry too much about performance. It is a polyfill afterall. For more performance input, lemme loop in @paullewis.

szager-chromium: The native implementation calculates intersections every frame, but it uses an idle callback with a timeout of 100ms to deliver them.  To match that, you would need to separate the code that calculates the intersection from the code that delivers the notifications.

I agree that calculating intersections in a scroll handler is problematic from a performance perspective.  It may very well trigger additional layouts, which is really anathema to whole purpose of this feature.

Even with the native implementation, notifications may be delayed by up to 100ms, so code that relies on up-to-the-frame information (e.g., input event handlers that want to validate the position of the clicked element before taking action) will need to use the takeRecords() method to get the most current information.  With that in mind, my suggestion is this:

Do the intersection calculations on a 100ms timer.  If takeRecords() is invoked, and too much time has passed since the last intersection calculation, then do the intersection calculation right away and return the up-to-date information to takeRecords().

mpb: Yes, the actual implementation fires on every rendered frame, basically. It can do this because it knows layout is clean at the time it runs, and doesn't have to worry about getBoundingClientRect triggering a layout.

The polyfill is closer to the actual implementation if it fires on every scroll and buffers, and triggers a callback based on a timer. But it will have performance implications.

surma: I did not know that rIC will be used for dispatch. Agreed then. I will adjust the code accordingly.

surma:
Review has been implemented. PTAL :)

triblondon:
I'm still concerned at the 100ms `setInterval` that is constantly polling the bounding rect of the element.  Was there a reason you can't use mutation observer for this?  If no elements on the page are changing I don't see why we need to keep polling, and it makes debugging a nightmare if your timeline is full of timer events.  With polyfills, performance of the feature being polyfilled is not a primary concern, but causing perf issues in other code certainly is. We want polyfills to make older browsers MORE reliable, not less.  On that basis, I would argue for a less aggressive, more conservative timer policy so that we reduce risk of crash or introducing jank, at the expense of a perfect reproduction of native perf on this specific feature.

We're also building on `getClientBoundingRect` here, which lacks width/height support in IE8 and below, so the polyfill is not likely to work in IE < 9, though I guess you could just use the innerWidth and innerHeight properties of the element if the boundingrect call returns no width or height.

I've started porting your tests into the polyfill.io library, and will take the polyfill once we've concluded this discussion.  Can you declare the licence terms under which you are publishing this code?  For polyfill.io we prefer MIT or CC0 but Apache (which appears to be one of the licences used in this repo) is fine too.

surma:
You are totally right. MutationObserver are a way better approach than `setTimeout`, will fix that. I would vote for not worrying about IE 8, 9 or 10 as they are officially deprecated and also don’t have MutationObserver.

triblondon:
In polyfill.io we care about IE 6+ currently (yes really), but I take a
pragmatic view of it. It seems like falling back to element width and
height is easy, so why not. No need to do any heroics to get support in old
browsers but this seems easy enough that you may as well do it.

surma:
Hats off to you for supporting IE6 :smile: 

As you suggested, I moved the implementation to use [MutationObserver](http://caniuse.com/#feat=mutationobserver) which makes this only work in IE11 and above, so there’s no point adding that fallback, or am I missing something?

triblondon:
Well, polyfill.io polyfills for MutationObserver, and we can make the IO polyfill depend on MO, then in browsers that don't have MO we ship that polyfill as well!

Also, we can offer some useful functionality here without MO, because many sites will only require the detection on scroll, so if you can shield the MO instantiation in a feature-detect, we can offer the polyfill on a wider range of browsers.

surma:
Fair enough. I’ll add the fallback dimensions :)

surma:
Done :)

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,231 @@
...
+  };
+
+  var debounce = function(fn, delay) {
+    var timer = null;
+    return function () {
+      if(timer) {
+        return;
+      }
+      var context = this, args = arguments;
+      clearTimeout(timer);
```
mpb: Won't `timer` always be `null` here?

surma: Not on rapid calls on the function being returned in line 213.

surma: Oh, nvm, you are right.

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,231 @@
...
+      }, delay);
+    };
+  };
+
+  var augmentRect = function(r) {
+    r.width = r.width || r.right - r.left;
+    r.height = r.height || r.bottom - r.top;
+    return r;
+  };
+})();
```
triblondon: Could we pass `this` into the IIFE here for safer global?

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,231 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+(function() {
+  // TODO: rootMargin are completely ignored for now
+
+  // Constructor
+  window.IntersectionObserver = function(callback, options) {
```
triblondon: Prefer to avoid assumed globals - bring `global` into the IIFE via an argument and use `global.IntersectionObserver` here.

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,231 @@
...
+    },
+
+    //
+    // Private API
+    //
+    _init: function() {
+      this._observationTargets = new Map();
+      this._boundUpdate = this._update.bind(this);
+      this.root.addEventListener('scroll', debounce(this._boundUpdate, 100));
+      this._mutationObserver = new MutationObserver(debounce(this._boundUpdate, 100));
```
triblondon: I think we want throttle here rather than debounce.  If the user scrolls smoothly and continuously down the page and the element we're watching transits the entire height of the viewport moving from out of view at the bottom to out of view at the top, the update fn will not be called at any point while the element is visible.

Also, there seems no reason to separately throttle the handlers for scroll and mutations, and finally the 100 is a magic number.  I'd suggest:

```javascript
this._boundUpdate = throttle(this._update.bind(this), POLL_INTERVAL);
this.root.addEventListener('scroll', this._boundUpdate);
this._mutationObserver = new MutationObserver(this._boundUpdate);
```

And then replace `debounce` with a minimal throttle implementation, such as:

```javascript
function throttle (fn, int) {
  var timer = null;
  return function () {
    if (timer) return;
    callback.apply(this, arguments);
    timer = setTimeout(function () {
      timer = null;
    }, int);
  };
}

File: polyfill/intersectionobserver-polyfill.js [hunk 8]
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    this._callback = callback;
+    this._root = options.root || null;
+    this._rootMargin = options.rootMargin || [0, 0, 0, 0];
+    this._thresholds = options.threshold || 0;
+    this._init();
+  };
+
+  window.IntersectionObserver.prototype = {
```
triblondon: Rather than these repeated references to global scope, it would be better to fully compose the `IntersectionObserver` object and then at the end assign it to the global scope, eg `global.IntersectionObserver = IntersectionObserver`.  This makes it easier to potentially add a UMD or otherwise change the way the functionality of the module is exported at a later date.

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,231 @@
...
+  // Constructor
+  window.IntersectionObserver = function(callback, options) {
+    options = options || {};
+
+    if(!(callback instanceof Function)) {
+      throw('callback needs to be a function');
+    }
+
+    if(options.root && !(options.root instanceof HTMLElement)) {
+      throw('Target needs to be an HTMLelement');
```
triblondon: Misleading error message.  "Target" => "Root"

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,231 @@
...
+      }
+
+      var root = this.root;
+      var ancestor = target.parentNode;
+      while (ancestor != root) {
+        if (!ancestor) {
+          throw('Target must be descendant of root');
+        }
+        ancestor = ancestor.parentNode;
+      }
```
triblondon: Could use `Node.prototype.contains()` here.
https://developer.mozilla.org/en/docs/Web/API/Node/contains

(polyfill.io has a polyfill for `contains`, though support is in any case very broad)

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,231 @@
...
+      this._observationTargets.set(target, {});
+    },
+
+    unobserve: function(target) {
+      this._observationTargets.delete(target);
+    },
+
+    disconnect: function() {
+      this._observationTargets.clear();
+      this.root.removeEventListener('scroll', this._boundUpdate);
```
triblondon: This doesn't work as currently written because the fun you are binding to the event handler is the one returned by debounce, not the original `boundUpdate` referenced here.  The throttle suggestion below addresses this issue though.

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,231 @@
...
+      this._mutationObserver.disconnect();
+      this._descheduleCallback();
+    },
+
+    takeRecords: function() {
+      this._update();
+      this._descheduleCallback();
+      var copy = this._queue.slice();
+      this._queue = [];
+      return copy;
```
triblondon: Or perhaps `return this._queue.splice(0);`?  Arguably what you already have is clearer.

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,231 @@
...
+      this._boundUpdate = this._update.bind(this);
+      this.root.addEventListener('scroll', debounce(this._boundUpdate, 100));
+      this._mutationObserver = new MutationObserver(debounce(this._boundUpdate, 100));
+      this._queue = [];
+    },
+
+    _update: function() {
+      var rootRect = this._rootRect();
+      this._observationTargets.forEach(function(oldIntersectionEntry, target) {
+        var targetRect = augmentRect(target.getBoundingClientRect());
```
triblondon: Personally I'd say abstracting the getBoundingClientRect call as well is probably clearer

```
getBoundingRect(target)

function getBoundingRect(el) {
    var r = el.getBoundingClientRect();
    r.width = r.width || r.right - r.left;
    r.height = r.height || r.bottom - r.top;
    return r;
 }
```

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,231 @@
...
+
+    _scheduleCallback: function() {
+      if(this._idleCallbackId) {
+        return;
+      }
+      this._idleCallbackId = window.requestIdleCallback(function() {
+        this._descheduleCallback();
+        this._callback(this._queue, this);
+        this._queue = [];
+      }.bind(this), {timeout: 100});
```
triblondon: Magic number, move to constant?

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,231 @@
...
+        this._scheduleCallback();
+        this._observationTargets.set(target, intersectionEntry);
+      }.bind(this));
+    },
+
+    _scheduleCallback: function() {
+      if(this._idleCallbackId) {
+        return;
+      }
+      this._idleCallbackId = window.requestIdleCallback(function() {
```
triblondon: Whooaaa.. requestIdleCallback is currently an ED that's only implemented in Chrome 47+, right? Can you use a standards track API here like raf or just plain setTimeout?

In the polyfill service we could drop in a simple requestIdleCallback polyfill and depend on that in the IO polyfill, but we prefer not to create dependency relationships on non-standard APIs (we're fine to ship non-standard API polyfills if requested explicitly by the user).  And anyone using this polyfill not via the polyfill service would be stuck.

triblondon:
@surma I am stuck in an airport for 6 hours so I did a more thorough code review.  Hope that's OK - and some of these are points you may well disagree with.

surma:
@triblondon Thank you very much for the thorough feedback. It’s highly appreciated :) Updated the code accordingly.

polyfill/intersectionobserver-polyfill.js
surma: Resolved with updating the definition of `_boundUpdate`.

polyfill/intersectionobserver-polyfill.js
surma: Sticking to my version in favor of readability.

triblondon:
Looks good. I'm working on getting the tests ported then I can test it in the polyfill service.

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,231 @@
...
+
+  scope.IntersectionObserver = IntersectionObserver;
+
+  var throttle = function(fn, int) {
+    var timer = null;
+    return function () {
+      if (timer) {
+        return;
+      }
+      callback.apply(this, arguments);
```
triblondon: `callback` => `fn` (sorry, this was wrong in my suggested code)

surma: I’m so lazy... fix’d :D

triblondon:
OK, I ported the tests and pushed them here:

https://github.com/Financial-Times/polyfill-service/blob/intersect-observer/polyfills/IntersectionObserver/tests.js

If you install and run the `intersect-observer` branch of the polyfill service, go to http://localhost:3000/test/tests?feature=IntersectionObserver&mode=all to run the tests in a local browser. Change `mode=all` to `mode=control` to run the tests without the polyfill.

Pretty good results in FF 38 and Chrome 49 using the polyfill, though it does not appear to support margins:

image

Not so good in Chrome 49 with experimental web platform flag enabled, testing the native support:

image

triblondon:
The problem with the native impl seems to be that records do not seem to be available synchronously with takeRecords().  Maybe.  What do you think?

triblondon:
Ah, the polyfill has a TODO against margin support, so I guess that my port of the tests is working then :-)  So it's mostly a case of why the native impl fails the tests.

surma:
Thanks a lot for your work, @triblondon. 

I’ll loop in @ojanvafai to check if my tests are wrong or the native implementation of Chrome is.

surma:
For the record: Talked to @ojanvafai and rest of the team. I am making assumptions in the tests that are not correct and break with the native implementation. I’ll fix that. @triblondon You might have to re-port some of the tests.

triblondon:
Would you consider using mocha/expect?  It seems to make for much neater tests, and you should be able to run the tests I made without the overhead of the polyfill service.  Just grab the [Mocha demo](https://mochajs.org/example/tests.html), add [Expect.js from CDNJS](https://cdnjs.cloudflare.com/ajax/libs/expect/1.16.0/expect.min.js) and replace the test cases with the contents of my file.

surma:
Will do.

surma:
Tests have been moved to mocha and I also made most of the tests properly asynchronous as `takeRecords()` isn’t very reliable for tests.

Some checks are commented out as the native implementation in Chrome 51 is missing the `intersectionRatio` field (will be fixed soon).

Other than that, all tests pass on both native and the polyfill implementation except for the margin tests. Margin tests fail on both, ironically, even on native. Looking into that.

szager-chromium:
@surma Maybe it will help you to refer to the root margin test for the native chromium implementation:

https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/LayoutTests/intersection-observer/root-margin.html&sq=package:chromium&type=cs

RByers:
Whoops, sorry - that must have been a side-effect of switching to the gh-pages branch (#120) sorry.  You'll probably need to update the PR to be based on gh-pages instead of the (now-deleted) master branch.

surma:
Continued in #121

PR #119, opened 2016-04-08
Modernising the example

jakearchibald:
Previous code used for-in on an array, which isn't recommended. Changed to for-of.


surma:
LGTM

PR #121, opened 2016-04-14
Add IntersectionObserver polyfill – Take #02

surma:
Continued from #116

### DIFF
--- a/polyfill/intersectionobserver-polyfill.js
+++ b/polyfill/intersectionobserver-polyfill.js
@@ -30,7 +30,7 @@ limitations under the License.
     this._callback = callback;
     this._root = options.root || null;
     this._rootMargin = options.rootMargin || [0, 0, 0, 0];
-    this._thresholds = options.threshold || 0;
+    this._thresholds = options.threshold || [0];
     this._init();
   };
 
@@ -47,11 +47,6 @@ limitations under the License.
     },
 
     get thresholds() {
-      // 0 means call callback on every change
-      // See note at http://rawgit.com/WICG/IntersectionObserver/master/index.html#intersection-observer-init
-      if(this._thresholds === 0) {
-        return 0;
-      }
       if(this._thresholds instanceof Array) {
         return this._thresholds;
       }
@@ -71,7 +66,7 @@ limitations under the License.
       if(!root.contains(target)) {
         throw('Target must be descendant of root');
       }
-      this._mutationObserver.observe(target, {
+      this._mutationObserver.observe(target.parentNode, {
         attributes: true,
         childList: true,
         characterData: true,
@@ -128,7 +123,8 @@ limitations under the License.
         var targetArea = targetRect.width * targetRect.height;
         var intersectionArea = intersectionRect.width * intersectionRect.height;
         var intersectionRatio = intersectionArea / targetArea;
-        if(!this._hasCrossedThreshold(oldIntersectionEntry.intersectionRatio, intersectionRatio)) {
+        console.log(intersectionRatio);
+        if(!this._hasCrossedThreshold(oldIntersectionEntry.intersectionRatio || 0, intersectionRatio)) {
           return;
         }
         var intersectionEntry = {
@@ -179,18 +175,22 @@ limitations under the License.
       };
     },
 
-    // FIXME: so hack, much performance, very JSON
     _hasCrossedThreshold: function(oldRatio, newRatio) {
-      if(this.thresholds === 0) {
-        return newRatio != oldRatio;
+      if(oldRatio === newRatio) {
+        return;
       }
+      var isOnBoundary = this.thresholds.map(function(threshold) {
+        return threshold == newRatio;
+      }).indexOf(true) !== -1;
+
       var b1 = this.thresholds.map(function(threshold) {
-        return threshold <= oldRatio;
+        return threshold < oldRatio;
       });
       var b2 = this.thresholds.map(function(threshold) {
-        return threshold <= newRatio;
+        return threshold < newRatio;
       });
-      return JSON.stringify(b1) !== JSON.stringify(b2);
+      var hasCrossedThreshold = b1.map((_, idx) => b1[idx] !== b2[idx]).indexOf(true) !== -1;
+      return isOnBoundary || hasCrossedThreshold;
     }
   };
 
@@ -233,6 +233,7 @@ limitations under the License.
 
   var getBoundingClientRect = function(el) {
     var r = el.getBoundingClientRect();
+    // Older IE
     r.width = r.width || r.right - r.left;
     r.height = r.height || r.bottom - r.top;
     return r;
--- a/polyfill/intersectionobserver-polyfill.js
+++ b/polyfill/intersectionobserver-polyfill.js
@@ -112,7 +112,7 @@ limitations under the License.
     _update: function() {
       var rootRect = this._rootRect();
       this._observationTargets.forEach(function(oldIntersectionEntry, target) {
-        var targetRect = target.getBoundingClientRect();
+        var targetRect = augmentRect(target.getBoundingClientRect());
         var intersectionRect = intersectRects(rootRect, targetRect);
         if(!intersectionRect) {
           return;
@@ -159,7 +159,7 @@ limitations under the License.
 
     _rootRect: function() {
       if(this._root) {
-        return this.root.getBoundingClientRect();
+        return augmentRect(this.root.getBoundingClientRect());
       }
       return {
         top: 0,
@@ -221,5 +221,11 @@ limitations under the License.
         fn.apply(context, args);
       }, delay);
     };
+  };
+
+  var augmentRect = function(r) {
+    r.width = r.width || r.right - r.left;
+    r.height = r.height || r.bottom - r.top;
+    return r;
   };
 })();
--- a/polyfill/intersectionobserver-polyfill.js
+++ b/polyfill/intersectionobserver-polyfill.js
@@ -1,0 +1,188 @@
+(function() {
+  // TODO: rootMargin are completely ignored for now
+  
+  // Constructor
+  window.IntersectionObserver = function(callback, options) {
+    options = options || {};
+    
+    if(!(callback instanceof Function)) {
+      throw('callback needs to be a function');
+    }
+    
+    this._callback = callback;
+    this._root = options.root || null;
+    this._rootMargin = options.rootMargin || [0, 0, 0, 0];
+    this._thresholds = options.threshold || 0;
+    this._init();
+  };
+
+  window.IntersectionObserver.prototype = {
+    //
+    // Public API
+    //
+    get root() {
+      return this._root || document;
+    },
+    
+    get rootMargin() {
+      return '0';
+    },
+    
+    get thresholds() {
+      // 0 means call callback on every change
+      // See note at http://rawgit.com/WICG/IntersectionObserver/master/index.html#intersection-observer-init
+      if(this._thresholds === 0) {
+        return 0;
+      }
+      if(this._thresholds instanceof Array) {
+        return this._thresholds;
+      }
+      return [this._thresholds];
+    },
+    
+    observe: function(target) {
+      if(!(target instanceof HTMLElement)) {
+        throw('Target needs to be an HTMLelement');
+      }
+      
+      var root = this.root;
+      var ancestor = target.parentNode;
+      while (ancestor != root) {
+        if (!ancestor) {
+          throw('Target must be decendant of root');
+        }
+        ancestor = ancestor.parentNode;   
+      }
+      
+      this._observationTargets.set(target, {});
+    },
+    
+    unobserve: function(target) {
+      this._observationTargets.delete(target);
+    },
+    
+    disconnect: function() {
+      this._observationTargets.clear();
+      this.root.removeEventListener('scroll', this._boundUpdate);
+      window.clearInterval(this._timeoutId);
+      this._descheduleCallback();
+    },
+    
+    takeRecords: function() {
+      this._update();
+      this._descheduleCallback();
+      var copy = this._queue.slice();
+      this._queue = [];
+      return copy;
+    },
+    
+    //
+    // Private API
+    //
+    _init: function() {
+      this._observationTargets = new Map();
+      this._boundUpdate = this._update.bind(this);
+      this.root.addEventListener('scroll', this._boundUpdate);
+      this._intervalId = window.setInterval(this._boundUpdate, 100);
+      this._queue = [];
+    },
+    
+    _update: function() {
+      var rootRect = this._rootRect();
+      this._observationTargets.forEach(function(oldIntersectionEntry, target) {
+        var targetRect = target.getBoundingClientRect();
+        var intersectionRect = intersectRects(rootRect, targetRect);
+        if(!intersectionRect) {
+          return;
+        }
+        var targetArea = targetRect.width * targetRect.height;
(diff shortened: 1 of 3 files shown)


surma:
Improved the intersection detection to match the native implementation (technically, my previous behavior was just straight up non-compliant with the spec). PTAL :)

triblondon:
Cool.  Is margin support coming, or should we ship a v1 at this point?

surma:
I vote for shipping a v1 without margin support.

jeremenichelli:
Hey there,

I just want to point out some stuff and do some questions around this polyfill.
- **line 120:** I'm seeing a return around `intersectionRect` that comes from a private function that always return non-empty object, so `!intersectionRect` will always be true making the `if` statement pointless.
- instead of this `scope.IntersectionObserver = scope.IntersectionObserver || IntersectionObserver` it would be better to return at the beggining of the closure if `intersectionObserver in window` is true so we don't execute unnecessary code when the API is present natively.
- wouldn't it be better to _throttle_ the action using `requestAnimationFrame`?
- I'm seeing some **ES2015** notation when the constructor prototype is declared, but also seeing some hack for old IE around `getBoundingClientRect` which is not consistent, the polyfill itself won't work in old IE because of this new object notation. I think it would be good to do it all in **ES5** or don't add hack for legacy browsers.
- Similar to the point from above, the code is asumming that `performance` is globally available, which is not in IE9, something like this https://gist.github.com/paulirish/5438650 should be added is the aim is to cover old IE, if not then makes no sense to add a `getBoundingClientRect` hack.

I hope this feedback is useful, you probably have answers and reasons for most of these, so sorry in advance if I missed something :grinning:

triblondon:
Personally I don't much care about relying on recent APIs or syntax because
syntax can be dealt with by transpiling and APIs by including further
polyfills. However, it does make sense to embed workarounds for browser
inconsistencies in older APIs where the API is widely supported but with
inconsistencies.

But it's really important to me to avoid embedding polyfills for things we
are not primarily trying to polyfill here. That's a slippery slope to
heavily duplicated front end code if a developer combines multiple
polyfills together.

So in summary:

* I'd favour being es5 compliant, I think that's a good practice for
polyfills
* we should not embed polyfills for other APIs which can be polyfilled
separately.
* it's ok to defensively code around inconsistencies in older APIs

jeremenichelli:
@triblondon I was actually thinking about that multi polyfilled scenario and I agree, that's a good point. I added three more points recently since I forgot them while adding the comment.

triblondon:
@surma where are you going to host this polyfill?  Is this WICG repo going to be the canonical source?  If so I can require it from here.  Ideally the polyfill would have its own repo though, and be published to npm.

surma:
I thought the WICG repo would be a good fit, but you are right in that it’s not really `npm install`-able this way. I have no strong opinion on this. @ojanvafai WDYT?

surma:
The ES2015 code was never meant to make it into the repo. Oversight on my part. Definitely was trying to use ES5 only.

I implement most of your remarks, @jeremenichelli. PTAL :)

jeremenichelli:
Looks great @surma, nice job :ok_hand:

jeremenichelli:
Just noticed one thing. Now that we're listening to the `resize` event, it should be removed inside the `disconnect` function https://github.com/surma-dump/IntersectionObserver/blob/981e7d355ef14069405e9131eca695be3cc5abfa/polyfill/intersectionobserver-polyfill.js#L88

surma:
Oh good catch! Fixed

polyfill/intersectionobserver_test.html
```diff
@@ -0,0 +1,253 @@
...
+          expect(records.length).to.be(1);
+          expect(records[0].intersectionRatio).to.be(1);
+          break;
+          case 1:
+          expect(records.length).to.be(1);
+          expect(records[0].intersectionRatio).to.be.lessThan(0.5);
+          break;
+          case 2:
+          expect(records.length).to.be(1);
+          expect(records[0].intersectionRatio).to.be.greaterTht dan(0.5);
```
triblondon: syntax error

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,250 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+// TODO: rootMargin are completely ignored for now
+(function(scope) {
+  if('IntersectionObserver' in window) {
+    return;
+  }
```
triblondon: Polyfill service prefers not to include feature tests inside the polyfill, because you may want to apply a polyfill over a native implementation in some cases, and detects are often too generous (this is actually a perfect case in point, because it will pass on the existing partial implementation in Chrome - though I realise it's behind a flag so not - yet - an issue in the wild)

triblondon:
OK, reported to polyfill service [here](https://github.com/Financial-Times/polyfill-service/compare/map...intersect-observer).  A few issues:

1. You haven't taken all the tests that I ported - it would be good to understand why these didn't make it:

- throws when root is not in the DOM
- throws when observed is not in the DOM
- throws if root does not contain child
- triggers if elements do not intersect when observing begins

2. Native implementation in Chrome 52 is failing the margins test, and also newly (as of 51 I think?) failing the first three of the tests above which I retained in my port.  Has the specced behaviour changed or were the tests (and the implementation) previously wrong?

3. The introduction of MutationObserver sets a floor of support at v11 of IE, which is a bit high. I don't have a MO polyfill in the polyfill service right now, and disabling the functionality provided by MO leaves us with a still reasonably useful polyfill which detects scroll-triggered intersections, and I'd rather have that than nothing at all.  So would you consider a feature gate on MO, ie `if ("MutationObserver" in window) ...`?  Sorry, I thought we had an MO polyfill I could use as a dependency, but we don't (yet).

4. The tests now require Promises and requestAnimationFrame, and that means I can't run them in our full set of legacy browsers (we start at IE6!).  You're going to hate this, but how about something like this (example for the margin test):

```
var marginChangeActions = [
  function() { childEl.style.top = "-15px"; },
  function() { childEl.style.top = "-14px"; },
  function() { childEl.style.top = "86px"; }
];
var doTest = function() {
  if (marginChangeActions.length === 0) done();
  var executeNextAction = marginChangeActions.shift();
  childEl.innerWidth;  // Force render as substitute for requestAnimationFrame
  executeNextAction();
}
io = new IntersectionObserver(
  function(records) {
    expect(records.length).to.be(1);
    doTest();
  },
  {root: rootEl, margin: "5px 10% 10% 15px"}
);
io.observe(childEl);
doTest();
```

Believe it or not this is also more compact than the current test code, doesn't involve running a tight-loop interval timer (mocha will handle timeouts for the test case overall), and it ought to run faster (because the current test code is adding an up-to-50ms latency on each action).  So I imagine I'm probably missing some fairly noddy reason why this is a stupid idea...

polyfill/intersectionobserver-polyfill.js
jeremenichelli: I agree with this, though I wouldn't like to overlap with a native implementation neither. In that case it would be necessary to see what is the partial part of the spec that's out and detect that and fix it like in `classList` API which Opera and IE10 doesn't support multiple class names or second argument in `toggle` method.

jeremenichelli:
Agree with @triblondon on the `MutationObserver` thing, apart from that here's a nice polyfill https://github.com/webcomponents/webcomponentsjs/blob/master/src/MutationObserver/MutationObserver.js

polyfill/intersectionobserver-polyfill.js
surma: I’ll can add an option to overwrite a native implementation, but I don’t think that should be the default behavior. I think the expected behavior is to do nothing if the browser has support for IO and polyfill it if it doesn’t.

polyfill/intersectionobserver_test.html
surma: How did this get in?!! o_O

surma:
1. Yes, the throws were not spec compliant. IOs are not supposed to throw in those cases, hence the removal of these tests. IO also should not trigger when elements do not intersect when starting to observe.
2. Native IO implemention got adjusted accordingly, as well.
3. I am not even sure MOs will stay as @ojanvafai pointed out that they might not actually cover all use-cases (think CSS animations), so I might have to revert to `setInterval` anyways. And if they stay, they are a dependency for this polyfill and must be polyfilled by the user if there’s no support for MOs. I need to start somewhere.
4. Also ties into point 3: I know your polyfill service starts at IE6, but this polyfill is for the web at it’s current state, not 2002. Both rAF and Promises can easily be polyfilled independently if needed. If it makes the code more readable, maybe. I’ll give it a try.

polyfill/intersectionobserver-polyfill.js
triblondon: OK, fair enough. It's not an important point and I accept the wider benefits.

triblondon:
Thanks.  I just think you're using unnecessarily modern APIs here - and we are trying to write a polyfill after all!  @jeremenichelli that looks like a good MO polyfill.  Want to add it to the polyfill service? :-)

jeremenichelli:
@triblondon not a problem at all, though I would contact Polymer team first since they are the authors of that polyfill to let them know you're going to add it to FT polyfill service. I think is a step in the right direction.

triblondon:
@surma If I refactor the tests for better legacy support can I push those changes to this PR?  I'm not a contributor currently - so I can also fork the repo and do a PR into your PR if that is better.  I understand why you might not want to bother doing that work yourself but I imagine you wouldn't object to it if the code is still clear and doesn't get any longer?

At the same time I'm looking into getting an MO polyfill into the polyfill service.

setInterval is a problem.  If you add setInterval into the source I'd remove it for my sites. In fact Alex Russell has had a go at me before for unwittingly serving a polyfill with a tight setInterval loop from the polyfill service.  I still have scars.

I don't believe perfect feature replication is a goal worth achieving at all costs - if you make the polyfill perfectly replicate the feature but at the same time make the browser juddery and unusable, or hard to debug, that's not a win.  The 99% use case for IO is objects coming into view on scroll, not as a result of CSS animation.  I asked around the dev teams here at the FT and they want margin support, and only care about intersections caused by scrolling. Obviously that's not a representative sample, and I'm just guessing that we are typical.

triblondon:
I suggested using IO to @afarkas who maintains [lazySizes](https://github.com/aFarkas/lazysizes) and got an [incredibly insightful response](https://github.com/aFarkas/lazysizes/issues/263#issuecomment-216067874), worth reading in full but he made some particular comments on the polyfill which I'll summarise:

* References to observed elements held strongly may create a memory leak
* Throttle should ideally use requestIdleCallback-like scheduling behaviour
* Scroll events within scrollable elements that are descendents of the main document may not be being captured but have the potential to bring observed elements into the viewport.

surma:
@triblondon I appreciate any time you spend on this. Feel free to contribute your code to this PR :)

If you want to ditch IO and add margin support, I am obviously fine with that. I just didn’t want to implement a CSS string parser and thought IO were a good choice. Apparently they aren’t. Also, I don’t want to be held responsible for @slightlyoff opening those old wounds 😉 

> I don't believe perfect feature replication is a goal worth achieving at all costs
Since I was actively ignoring margins all this time and is perfectly fair. If no one is veto-ing your assessment of the 99% use-case, go for it :) 

Regarding the feedback from @afarkas:

* `WeakMap` seems like a good idea. Not sure what needs to change in terms of implementation
* Currently, `setTimeout(..., 100)` is used to deliver the data to the callbacks. I thought that was pretty rIC-like.
* That should only be a problem if the scroll event is actively stopped from propagating/bubbling, right? Or am I missing something? Personally, I’d be fine with ignoring that circumstance.

triblondon:
Scroll doesn't bubble.

I've updated my tests.  I think this pattern works really well, offers a more linear read of the tests and assertions, tests run faster, and they're compatible with older browsers.  It's actually slightly shorter too. I think your margin tests were slightly awry because the test case does not set up the IO with any margins!  So I reset that test case to the steps I had in my last work on it.

I'm now finding that Chrome 52 (canary) native passes all except the margin tests:

image

Chrome 50 (stable) using the polyfill:

image

I don't have push on either your fork or the WICG, so I can make a new surma-dump fork if you'd like these changes proposed for merge into your PR, but in the meantime I pushed them to https://github.com/Financial-Times/polyfill-service/compare/intersect-observer#diff-c0d2dbf8759d522d6028477a3a43497f

surma:
Great! Awesome stuff, @triblondon. Made you collab on my fork. So feel free to push your changes. Looking forward to it.

triblondon:
OK, I pushed.  Notes in the commit messages should be fairly explanatory.  I guess your next steps are to investigate the difference in native vs polyfill test results, and look at @afarkas's points?

triblondon:
@surma will you be picking this back up at some point?  We have a [PR open in the polyfill service](https://github.com/Financial-Times/polyfill-service/pull/684) to merge this in once its ready.

surma:
Sorry for the absence, but Google I/O was happening and was quite the marathon. Just got back and will continue working on this soon. Thanks for your work and for the patience :)

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,250 @@
...
+    };
+  };
+
+  var getBoundingClientRect = function(el) {
+    var r = el.getBoundingClientRect();
+    if(!r) {
+      return null;
+    }
+    // Older IE
+    r.width = r.width || r.right - r.left;
```
birkir: Safari throws an read-only error on setting this property. I fixed it by checking if the property is writable using the property descriptor getter.

```javascript
var desc = Object.getOwnPropertyDescriptor(r, 'width');
if (desc && desc.writable) {
  r.width = r.width || r.right - r.left;
  r.height = r.height || r.bottom - r.top;
}
```

polyfill/intersectionobserver-polyfill.js
```diff
@@ -0,0 +1,250 @@
...
+      if(this._observationTargets.has(target)) {
+        return;
+      }
+      if(!(target instanceof HTMLElement)) {
+        throw('Target needs to be an HTMLelement');
+      }
+
+      var root = this.root;
+      var ancestor = target.parentNode;
+      if(!root.contains(target)) {
```
birkir: IE11 does not have the `contains` method, it needs to be polyfilled if wanting that browser support.

polyfill/intersectionobserver-polyfill.js
jeremenichelli: I have this problem when using `getBoundingClientRect`, I think the best way to get around this is to assign it to return a different object.

```js
var r = el.getBoundingClientRect();
  if(!r) {
    return null;
  }
  return {
    top: r.top
    // the rest of the metrics...
    width: r.width || r.right - r.left;
    height: r.height || r.bottom - r.top;
  }
```

Extra code, but safe.

polyfill/intersectionobserver-polyfill.js
```diff
@@ -58,6 +58,12 @@
         ancestor = ancestor.parentNode;
       }
 
+      this._mutationObserver.observe(target, {
+        attributes: true,
+        childList: true,
+        characterData: true,
```
cpeterso: Do you need to observe `characterData` changes to track the target element's visibility? Can you use `attributeFilter` to observe only changes in the target's dimension attributes?

surma:
Continued in #135

PR #125, opened 2016-04-29
mismatched closing tag

atotic:
</h3> -> </h4>


ojanvafai:
Interesting. I think this is auto-generated. @tabatkins @szager-chromium right? Is this a bug in bikeshed?

mpb:
the .bs file is the input file bikeshed uses to generate its html.

PR #130, opened 2016-05-20
Specify order of entries sent to callback.

szager-chromium:
Specify that changes are queued according to the order in which observe() was called on their target Elements.


szager-chromium:
Closing this because it pulls in a bunch of unrelated changes, presumably due to bikeshed version mismatch.

PR #131, opened 2016-05-20
Change order

szager-chromium:
Specify the order in which targets are processed, and consequently the order of change entries sent to the observer callback.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -583,7 +583,7 @@ To <dfn>run the update intersection observations steps</dfn> for an
 	whose {{IntersectionObserver/root}} is in the DOM tree of a document in |unit|.
 3. For each |observer| in |observer list|:
 	1. Let |rootBounds| be |observer|'s <a>root intersection rectangle</a>.
-	2. For each |target| in |observer|'s internal {{[[ObservationTargets]]}} slot:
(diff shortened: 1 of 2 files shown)


plehegar:
@szager-chromium , Can you join WICG using
https://www.w3.org/community/wp-login.php?redirect_to=%2Fcommunity%2Fwicg%2Fjoin ? 
This will allow me to clear the ipr check. Thank you in advance.

szager-chromium:
@plehegar, I went through the sign-up process for W3C, but because my affiliation is with Google, they require that I create the account with my @google.com email address.  If I do so, will I still be able to link my github account, which is @chromium.org, to my W3C account?

plehegar:
@szager-chromium , yes certainly. The linkage doesn't mind about email addresses. Btw, if I remember correctly, you may even still be able to change your email address on your W3C account once it got associated with your company.

szager-chromium:
@plehegar, OK should be all set now.

plehegar:
@szager-chromium , almost there. You need to join WICG by using 
 https://www.w3.org/community/wp-login.php?redirect_to=%2Fcommunity%2Fwicg%2Fjoin

PR #150, opened 2016-08-16
Use native stuff 

shvaikalesh:
1. `Node.prototype.contains` is available since IE5: replace custom helper with it.
2. `Array.isArray` should be available, because `es5shim` is required.
3. `use strict` is cool (explicit error stuff), and without it crankshaft will bailout on parameter reassignment.
4. They say `bind` is slow, lets pass `this` to es5 array methods as last argument.
5. Replace `filter` with `indexOf` + `splice`, should be faster & safer than reassignment.

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -186,7 +186,7 @@ IntersectionObserver.prototype.takeRecords = function() {
  */
 IntersectionObserver.prototype._initThresholds = function(opt_threshold) {
   var threshold = opt_threshold || [0];
-  if (!isArray(threshold)) threshold = [threshold];
+  if (!Array.isArray(threshold)) threshold = [threshold];
 
   return threshold.sort().filter(function(t, i, a) {
     if (typeof t != 'number' || isNaN(t) || t < 0 || t > 1) {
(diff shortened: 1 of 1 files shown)


philipwalton:
I guess these make sense. Originally I was going for wider browser support, but we're already asking IE7-8 users to include an ES5 polyfill, so I think we can safely assume these features will be supported.

shvaikalesh:
Please, don't merge yet, I will test `contains` on `document` in IE7-8.

shvaikalesh:
IE stuff is fixed by https://github.com/WICG/IntersectionObserver/pull/150/commits/804107068870491c706f484d188f5fab53357acf.

philipwalton:
Thanks, I think you need to request to join the WICG and sign the CLA before I can merge this:
https://www.w3.org/community/wicg/

I'm not 100% on the process though, so I'll see if I can find out more, but as it stands the checks fail because it doesn't recognize your username.

shvaikalesh:
I have signed it.

marcoscaceres:
@shvaikalesh thanks for signing it 👍

philipwalton:
Thanks!

shvaikalesh:
@philipwalton thanks for the merge. Very nice code, btw 👍

Issue #69, opened 2015-11-05
How should zero-dimension targets be handled?


szager-chromium:
This is probably a legitimate use case, but it's technically a divide-by-zero error when calculating the visible area ratio.

I think common sense dictates that visible area ratio should be 0% if the target location does not intersect the root, or 100% if it does.

ojanvafai:
Decided to change the algorithm to expand 0 width/height dimensions of intersection rects by 1 for the purposes of computing visibility. The rect returned to JS will still be 0 width/height.

If the rect doesn't not intersect in then the intersectionRect will be null and it will be considered not intersecting.

ojanvafai:
This needs reworking. If there's a 0x5px rect at the bottom, we'd never notify intersection when you hit the bottom of the scroller. So we should revisit this.

ojanvafai:
Might need a bool entry on the change record visibile/not-visible.

ojanvafai:
```
<div id=root>
  ... lots of content
  <div id="target" style="height: 0px; width: 100%">
</div>
```

Adding a pixel to the target size doesn't work because it makes it stick out the bottom.

ojanvafai:
We think the spec should change to just invoke an inclusive intersection (includes the boundary points in the intersection). If implementations would like to do a different implementation with the same behavior (e.g. through inflating rects), they are welcome to.

We should make sure to show a picture of what intersect rects you get back in these cases. The intersection rect in the change record can be zero-sized.

rjgotten:
> We think the spec should change to just invoke an inclusive intersection (includes the boundary points in the intersection).

That leads to there being no way to discern an intersection with a coincident edge or coincident corner point from there being no intersection at all. With a default threshold value of `[0]` it means you can no longer reliably determine in what direction the threshold was crossed.

I've already had to switch back from Chrome's native implementation (which seems to use inclusive intersection) to using a polyfill implementation which uses exclusive intersection and completely ignoring the existence of the native implementation, because of this problem.

szager-chromium:
This seems to be a common stumbling block for developers.  You can work around this by keeping track of the previous intersection event for a target:

var wasIntersecting = false;
function intersectionCallback(changes) {
  changes.forEach(function (change) {
    var isIntersecting = !wasIntersecting;
    wasIntersecting = isIntersecting;
    if (isIntersecting)
      doIntersectingThing();
    else
      doNotIntersectingThing();
  });
}

[...]

PR #155, opened 2016-09-20
Handle elements set to display:none

### LINKED ISSUE #145: The polyfill reports intersections for elements set to `display: none`
Externally reported here: https://github.com/Financial-Times/polyfill-service/issues/776

An easy fix would be to ignore `display: none` elements, but I'm wondering if spec clarification is needed here. Do we also need to consider how we handle intersecting but otherwise invisible elements (e.g. `opacity: 0` or `visibility: hidden`)?

Note that this is happening in the polyfill because a `display: none` element will report a `(0, 0, 0, 0)` rect, which (if it were visible) _would_ be considered an intersection as discussed in #69.

philipwalton:
This pull request fixes #145. It also fixes a related bug where elements with a rect of (0, 0, 0, 0) in the top-left corner of the page don't trigger an intersection.

To accomplish this, I added an `isIntersecting` flag to the `IntersectionObserverEntry` object as discussed in #69. I assume this is OK since it sounds like there's support to add it to the spec. If not, I can refactor the logic, but having this flag greatly simplifies things.

### DIFF
--- a/polyfill/intersection-observer-test.js
+++ b/polyfill/intersection-observer-test.js
@@ -565,6 +565,47 @@ describe('IntersectionObserver', function() {
     });
 
 
+    it('handles elements with display set to none', function(done) {
+
+      var spy = sinon.spy();
+      io = new IntersectionObserver(spy);
+
+      runSequence([
+        function(done) {
+          rootEl.style.position = 'absolute';
+          rootEl.style.top = '0px';
+          rootEl.style.left = '0px';
+          rootEl.style.width = '0px';
+          rootEl.style.height = '0px';
+          rootEl.style.display = 'none';
+          io.observe(rootEl);
+          setTimeout(function() {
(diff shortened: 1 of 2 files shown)


philipwalton:
Ping @szager-chromium @ojanvafai can one of you PTAL so I can release these changes?

triblondon:
@philipwalton are you still intending to ship this?  It blocks our issue https://github.com/Financial-Times/polyfill-service/issues/776 and I'd love to pull the fix.

philipwalton:
@triblondon yes, I'm here with @szager-chromium at the CSS WG F2F and we're looking at it now. We should be able to merge. Sorry for the delay.

philipwalton:
@triblondon [v0.2.0](https://www.npmjs.com/package/intersection-observer) has been published to npm. Let me know if you need anything else to get unblocked.

Issue #159, opened 2016-09-30
Clarify zero-pixel intersection behavior


mstange:
I'm confused by this test:

``` js
    it('handles targets on the boundary of root', function(done) {

      var spy = sinon.spy();
      io = new IntersectionObserver(spy, {root: rootEl});

      runSequence([
        function(done) {
          targetEl1.style.top = '0px';
          targetEl1.style.left = '-21px';
          targetEl2.style.top = '-20px';
          targetEl2.style.left = '0px';
          io.observe(targetEl1);
          io.observe(targetEl2);
          setTimeout(function() {
            expect(spy.callCount).to.be(1);
            var records = sortRecords(spy.lastCall.args[0]);
            expect(records.length).to.be(1);
            expect(records[0].intersectionRatio).to.be(0);
            expect(records[0].target).to.be(targetEl2);
            done();
          }, ASYNC_TIMEOUT);
},
```

Neither `targetEl1` nor `targetEl2` contain any pixels that intersect with `rootEl`. Why does the test expect an intersection observer callback for `targetEl2`?

cc @tobytailor

mstange:
In particular, the spec says:

> callback will be invoked when intersectionRect’s area changes from greater than or equal to any threshold to less than that threshold, and vice versa.

and I don't understand how this justifies treating the two elements differently.

philipwalton:
FYI, the discussion around this behavior is captured in #69.

mstange:
Well, somewhat... as far as I can tell, the discussion does not mention requiring browsers to fire the callback when the element becomes adjacent to an edge. Even an `isIntersecting` flag would be false for both `targetEl1` and `targetEl2`, wouldn't it?

philipwalton:
This [comment](https://github.com/WICG/IntersectionObserver/issues/69#issuecomment-180128037) in that thread describes the updated behavior:

> We think the spec should change to just invoke an inclusive intersection (includes the boundary points in the intersection).

In the test case you referenced above, `targetEl2` is on the boundary (though not overlapping) and `targetEl1` is not on the boundary (and also not overlapping). The use of inclusive intersections explains why an intersection is expected for `targetEl2`.

mstange:
Ah, thank you. It wasn't clear to me whether that was just one idea of many, or an actual decision.
I'd appreciate the spec being updated, so that the tests can actually test specced behavior.

Issue #156, opened 2016-09-21
Point 8 in section 3.2.5 is wrong


tobytailor:
"Otherwise, let threshold be the index of the first entry in observer.thresholds whose value is greater than intersectionRatio." should be "Otherwise, let threshold be the index of the last entry in observer.thresholds whose value is smaller than intersectionRatio."

ziyunfei:
No, it should be "let threshold be the index of the last entry in observer.thresholds whose value is smaller or equal than intersectionRatio; let threshold = threshold + 1"

tobytailor:
Even better :)

szager-chromium:
Good point, thanks.

PR #180, opened 2016-11-19
Be more precise in handling of thresholds.

szager-chromium:
In particular, describe how zero-area intersections are handled.

Issue #156 
Issue #159

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -597,15 +597,12 @@ To <dfn>run the update intersection observations steps</dfn> for a
 			algorithm on |target|.
 		5. Let |targetArea| be |targetRect|'s area.
 		6. Let |intersectionArea| be |intersectionRect|'s area.
-		7. Let |intersectionRatio| be |intersectionArea| divided by |targetArea|
-			if |targetArea| is non-zero,
-			and <code>0</code> otherwise.
-		8. Let |doesIntersect| be true if |targetRect| and |rootBounds| intersect or are edge-adjacent,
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -600,25 +600,26 @@ To <dfn>run the update intersection observations steps</dfn> for a
...
-			If |intersectionRatio| is equal to <code>1</code>,
-			let |threshold| be the length of |observer|.{{thresholds}}.
-			Otherwise, let |threshold| be the index of the first entry in |observer|.{{thresholds}}
-			whose value is greater than |intersectionRatio|.
-		9. Let |intersectionObserverRegistration| be the {{IntersectionObserverRegistration}} record 
+		8. Let |doesIntersect| be true if |targetRect| and |rootBounds| intersect or are edge-adjacent,
+			even if the intersection has zero area (because |rootBounds| or |targetRect| have
+			zero area); otherwise, let |doesIntersect| be false.
+		9. If |doesIntersect| is false, let |thresholdIndex| be <code>0</code>.<br>
+			Otherwise, if |targetRect| has zero area, let |thresholdIndex| be <code>1</code>.<br>
```
ojanvafai: Instead of 1, should this be "the length of |observer|.{{thresholds}}"? I know our current code does 1, but we had talked of exposing the threshold index on the IntersectionObserverEntry. I don't know if we should do that, but if we did, then we would probably want to change this.

Issue #149, opened 2016-08-15
Should post a task instead of using idle callbacks


esprehn:
Idle callbacks brings a bunch of baggage:
- Idle callbacks means on busy machines (or situations where the scheduler thinks the work is not a priority) that the callback is delayed 100ms.
- The 100ms delay is specifically targeting a specific ads standard, but this API is more general than that. Different timeouts make sense for different use cases.
- Having to use takeRecords + a scroll listener means you're listening for every scroll even when it's not needed which breaks one of the big use cases which was not avoid scroll listeners.
- Calling requestIdleCallback from inside an IO callback always pushes you a frame forward because of the semantics of rIc.
- Only Chrome supports idle callbacks so far and the scheduler work to have them is non-trivial. Coupling a super useful feature like IO to it is unfortunate for other implementations.

We've also never forced this kind of scheduling with timeouts on the platform before. I don't think we understand the ecosystem implications properly, and some authors are already not happy with it.

We'd be better of removing this from the spec and instead letting developers compose the primitives themselves (rIc, IO, setTimeout, etc.).

Long term we can consider sugar to make idle callbacks nicer in a general purpose scheduling API, for example:

``` js
Scheduler.atIdle(Function, opts) => Function
Scheduler.atTimeout(Function, opts) => Function
Scheduler.atMicrotask(Function, opts) => Function
```

etc. which allows you to easily vend a function that defers the execution to a different point in time, then you can do:

``` js
var observer = new IntersectionObserver(Scheduler.atIdle((records) => {
  // ...
}, {timeout: 100}));
```

ziyunfei:
http://crrev.com/31768c865c398d024d8adad8cd5d9f2b182d12c2

ojanvafai:
I'll try to update the spec shortly. Didn't seem urgent since no other
browser implements idle callbacks.

PR #182, opened 2016-11-19
Switch from idle callback to post task.

szager-chromium:
Issue #149


ojanvafai:
This looks good to merge, but has a conflict now after the other merges.

Issue #165, opened 2016-10-25
need notifications when observe is called on offscreen elements


ojanvafai:
Right now, we assume that an element is not visible when observe is called, queue up a run of the rendering pipeline, and then a callback gets delivered if the element turns out to be visible. There are use cases where the author content wants to assume the opposite, that the element is visible and to get a callback if it's hidden.

There's a couple options to solve this:
1. Add a ternary option to the constructor. assume:visible|hidden|both or something like that.
2. Just always have it fire the callback once whenever observe is called.
#2 seems simplest. I suspect the overhead of delivering the callbacks once to not be too much.

@szager-chromium @esprehn

szager-chromium:
I think #2 makes sense.

ziyunfei:
https://bugzilla.mozilla.org/show_bug.cgi?id=1335644

jetvillegas:
The danger here is noted in the first comment on the Firefox bug above ^

Scripts that lazy-load on notification may fire prematurely based on that first notification. I'm inclined not to fix it in Firefox since this is an Ad-perf feature and the lazy-load issue will not be a good effect. Since Chrome is shipping it this way now, can you cite public cases that count on this behavior?

tantek:
Indeed, assuming the use-case referred to in the issue description, presumably those users are now using this? Could you (@ojanvafai) provide URLs to actual examples that are using the Chrome implementation? Or if they're waiting on other implementations that would be good to know too.

szager-chromium:
I'm trying to remember where the original request came from, but I believe the use case was that someone needed a "one-shot" observer, something along the lines of:

var newElement = foo();
var observer = new IntersectionObserver(function(changes) {
  doSomethingInterestingBasedOnInitialVisibility(newElement, changes.pop().isIntersecting);
  observer.unobserve(newElement);
});
observer.observe(newElement);

I will chase down Ojan and see if he remembers more, or where the request came from.

a2sheppy:
What about if you're constructing a page which is performing tasks when elements become visible or not-visible, and you don't know whether an element is visible at the time it's added to the page or not? Getting this notification after observe() would let you easily handle that initial state properly.

szager-chromium:
Right -- without the initial notification, there's no way to tell definitively that a newly-added element is not visible, unless you use some gymnastics like this:

var newElement = foo();
var gotNotification = false;
var observer = new IntersectionObserver(function(changes) { gotNotification = true; });
[...]

PR #192, opened 2017-01-20
Always send a notification on the first frame after observe().

szager-chromium:
issue #165


ojanvafai:
Lol. Such a tiny change!

PR #200, opened 2017-03-06
Use strict equal

1000ch:
Slightly faster.


1000ch:
Marked as non-substantive for IPR from ash-nazg.

philipwalton:
Thanks, but the double equals is in accordance with the style guide.

PR #204, opened 2017-03-27
zero-size targets should have intersectionRatio 1 according to spec

sillero:
as stated on https://github.com/WICG/IntersectionObserver/issues/190 and https://wicg.github.io/IntersectionObserver/#dom-intersectionobserverentry-intersectionratio

*If the boundingClientRect has non-zero area, this will be the ratio of intersectionRect area to boundingClientRect area. **Otherwise, this will be 1 if the isIntersecting is true, and 0 if not.***

Note: I already applied to join the group

### DIFF
--- a/polyfill/intersection-observer-test.js
+++ b/polyfill/intersection-observer-test.js
@@ -553,7 +553,7 @@ describe('IntersectionObserver', function() {
 
       io = new IntersectionObserver(function(records) {
         expect(records.length).to.be(1);
-        expect(records[0].intersectionRatio).to.be(0);
+        expect(records[0].intersectionRatio).to.be(1);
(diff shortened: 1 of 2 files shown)


sillero:
Joined the group and linked my w3 account to github. Not sure what else to do.

yoavweiss:
@sillero - Thanks for joining!

@dontcallmedom - I've added @sillero to the system and to the WICG group there, but it still claims they're not in the right groups. Can you take a look?

yoavweiss:
@dontcallmedom - nm, had issues with the W3C ID. Fixed now.

philipwalton:
Thanks!

PR #205, opened 2017-03-31
getBoundingClientRect should always return a rect

robwalch:
I'm running into issues with my projects tests running on IE11 in Windows 7.

IE11 throws an "Unspecified error." at the line that calls `getBoundingClientRect`. Catching the exception prevents the error from breaking our tests. Returning an empty rect prevents the polyfill from creating an entry with a `boundingClientRect` of `undefined` which would throw in `_checkForIntersections`.

This makes the polyfill work to spec in IE11 by working around this issue https://connect.microsoft.com/IE/feedback/details/829392/calling-getboundingclientrect-on-an-html-element-that-has-not-been-added-to-the-dom-causes-unspecified-error

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -623,8 +623,15 @@ function computeRectIntersection(rect1, rect2) {
  * @return {Object} The (possibly shimmed) rect of the element.
  */
 function getBoundingClientRect(el) {
-  var rect = el.getBoundingClientRect();
-  if (!rect) return;
+  var rect;
+
+  try {
+      rect = el.getBoundingClientRect();
+  } catch (e) {/* ignore Windows 7 IE11 "Unspecified error" */}
+
+  if (!rect) {
+      return getEmptyRect();
+  }
 
   // Older IE
   if (!rect.width || !rect.height) {


yoavweiss:
@robwalch - can you join the WICG in order to appease the IPR bots? Thanks! :)

robwalch:
Thanks @yoavweiss. Will this do https://www.w3.org/users/98675?

yoavweiss:
@robwalch Can you also go to https://www.w3.org/community/wicg/participants and join the CG? Thanks!

robwalch:
@yoavweiss I joined the group. Not sure what I need to do appease the PR bots now
 Thanks!

yoavweiss:
And the bots are appeased! :)

amatiasq:
Any updates here? We've found this bug too
Will it be included on the next release?

philipwalton:
I'm happy to merge this, but I'm curious why the current test suite wasn't detecting this issue. The tests run in IE11 and there's a test for elements not yet in the DOM.

@robwalch can you take a look and see if the tests need to be updated so future changes don't regress with regards to this?

robwalch:
@philipwalton my guess is that's because the target elements are created as part of the test fixture, and for the test in question, only removed right before running the test.

I'll add a test that uses a new element that has never been added to the DOM:

`targetElN =  document.createElement('div');`

That does the trick in IE11's dev console.

robwalch:
@philipwalton I added the test here https://github.com/WICG/IntersectionObserver/pull/219 so we could see it fail and hopefully fixed with the changes here.

amatiasq:
Is this in the new version?

philipwalton:
Yes, if you look at this history you'll see it was merged.

amatiasq:
Yes I did, the question is if what is merged is on the last version.

philipwalton:
https://github.com/WICG/IntersectionObserver/commits/gh-pages

PR #219, opened 2017-06-08
Add a test for target elements not yet added to the DOM

robwalch:
This adds a test which fails in IE 11 on Windows 7 with this bug https://connect.microsoft.com/IE/feedback/details/829392/calling-getboundingclientrect-on-an-html-element-that-has-not-been-added-to-the-dom-causes-unspecified-error

This issue was not caught before, because all elements tested are added with the test fixtures. The `getBoundingClientRect` bug in IE 11 is specific to elements created using `document.createElement()` that have not yet been added to the DOM.

This should be fixed with #205

### DIFF
--- a/polyfill/intersection-observer-test.js
+++ b/polyfill/intersection-observer-test.js
@@ -609,7 +609,72 @@ describe('IntersectionObserver', function() {
     });
 
 
-    it('handles root/target elements not yet in the DOM', function(done) {
+    it('handles target elements not yet added to the DOM', function(done) {
+      var spy = sinon.spy();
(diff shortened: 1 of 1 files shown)


philipwalton:
Thanks! Can you rebase against gh-pages to pick up the latest changes? It looks like your test is failing because it's making some old assumptions.

robwalch:
@philipwalton I rebased against gh-pages. This branch only contains a new test "handles target elements not yet added to the DOM" that should fail in IE. The fix is in https://github.com/WICG/IntersectionObserver/pull/205

philipwalton:
Thanks!

PR #224, opened 2017-06-28
Add minimal isIntersecting polyfill for Edge 15

### LINKED ISSUE #211: Extend feature detection with isIntersecting
👋  Hello.

I discovered that according to [caniuse](http://caniuse.com/#feat=intersectionobserver) Edge 15 supports the Intersection Observer API. While playing around with it I discovered though, that Edge does not implement the [`isIntersecting` property](https://wicg.github.io/IntersectionObserver/#dom-intersectionobserverentry-isintersecting).

The Firefox implementation (behind a flag currently) is also missing this property on `IntersectionObserverEntry`.

Would it make sense to extend the [feature detection](https://github.com/WICG/IntersectionObserver/blob/gh-pages/polyfill/intersection-observer.js#L23-L27) which is currently:

```
if ('IntersectionObserver' in window &&
    'IntersectionObserverEntry' in window &&
    'intersectionRatio' in window.IntersectionObserverEntry.prototype) {
  return;
}
```
with a check for this property? I think it's really handy and this property missing could lead to "unexpected behavior" on developer's side.

```
if ('IntersectionObserver' in window &&
    'IntersectionObserverEntry' in window &&
    'intersectionRatio' in window.IntersectionObserverEntry.prototype &&
    'isIntersecting' in window.IntersectionObserverEntry.prototype) {
  return;
}
```

Thanks in advance. :)

nolanlawson:
Fixes #211.

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -22,7 +22,18 @@
 // features are natively supported.
 if ('IntersectionObserver' in window &&
     'IntersectionObserverEntry' in window &&
-    'intersectionRatio' in window.IntersectionObserverEntry.prototype) {
+    'intersectionRatio' in IntersectionObserverEntry.prototype) {
+
+  // Minimal polyfill for Edge 15's lack of `isIntersecting`
+  // See: https://github.com/WICG/IntersectionObserver/issues/211
+  if (!('isIntersecting' in IntersectionObserverEntry.prototype)) {
+    Object.defineProperty(IntersectionObserverEntry.prototype,
+      'isIntersecting', {
+      get: function () {
+        return this.intersectionRatio > 0;
+      }
+    });
+  }
   return;
 }
 


nolanlawson:
Hm, I tried to fix the IPR issue but I'm getting a "forbidden" error from labs.w3.org when I try to revalidate. :confused:

philipwalton:
Hmmm, I'm not sure why. @marcoscaceres any ideas?

marcoscaceres:
@nolanlawson it seems to think that you didn't join the WICG? Can you confirmed that you joined? If not, you can join here: https://www.w3.org/community/wicg/

nolanlawson:
My bad; I hadn't joined the WICG yet. Just sent a request for membership.

marcoscaceres:
np. Let me know once that's gone through so I can pull the right levers on this end.

nolanlawson:
OK, I'm in the WICG now. :)

philipwalton:
Thanks!

philipwalton:
Published version 0.4.0 with this change.

pocketjoso:
Hi everyone, thanks for the work on this.

I'm confused about this change, it seems inconsistent with how `isIntersecting` is documented at least on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API):

>  You can tell which direction the visibility changed in (that is, whether the element became more visible or less visible) by checking the value of the isIntersecting property on the IntersectionObserverEntry passed into the callback function at the time of the visibility change. If isIntersecting is true, the target element has become at least as visible as the threshold that was passed. If it's false, the target is no longer as visible as the given threshold.

The `isIntersecting` polyfill introduced in this PR doesn't match this behavior, which is problematic. Is it MDN's docs that are incorrect, or is this `isIntersect` polyfill too simple?

nolanlawson:
The MDN document is still correct. `isIntersecting` describes the _direction_ that the visibility changed. The issue is that Edge 15 implements an older version of the spec, and this polyfill brings it as close as possible to the current version.

Have you tested Edge 15 and seen any issues with this behavior? In my own testing, it's working pretty well – `isIntersecting` correctly flips from true to false or vice versa when the boundary is passed.

pocketjoso:
@nolanlawson I do think it will work well enough for most use cases; if you're only interested in things like knowing whether something came into view (started intersecting) or not. However if you want to do things like react to when something starts going out of view, with this implementation it's not possible to do using the `isIntersecting` value..

PR #229, opened 2017-07-07
fix rootMargin value for Firefox

midzer:
Hi,

Firefox ESR 52.2.0 gave a helpful error message in console about malformed `rootMargin` value when using this polyfill.

With this fix error message goes away and polyfill is working smoothly :)

Have a nice weekend
midzer


philipwalton:
Marked as non-substantive for IPR from ash-nazg.

philipwalton:
Thanks!

PR #231, opened 2017-07-20
Handle display:none on target parents

### LINKED ISSUE #230: Polyfill: Elements with ancestors with display: none should not intersect
See #145, #155 

Fixed was the issue that elements with `display: none` should not intersect:

```js
IntersectionObserver.prototype._computeTargetAndRootIntersection =
    function(target, rootRect) {

  // If the element isn't displayed, an intersection can't happen.
  if (window.getComputedStyle(target).display == 'none') return;
...
```

However elements with **ancestors** with `display: none` are still reported as intersecting.

philipwalton:
Fixes #230.

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -368,6 +368,10 @@ IntersectionObserver.prototype._computeTargetAndRootIntersection =
 
   while (!atRoot) {
     var parentRect = null;
+    var parentComputedStyle = window.getComputedStyle(parent);
+
+    // If the parent isn't displayed, an intersection can't happen.
+    if (parentComputedStyle.display == 'none') return;
 
     // If we're at the root element, set parentRect to the already
     // calculated rootRect. And since <body> and <html> cannot be clipped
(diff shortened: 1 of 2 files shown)


polyfill/intersection-observer.js
```diff
@@ -368,6 +368,10 @@ IntersectionObserver.prototype._computeTargetAndRootIntersection =
 
   while (!atRoot) {
     var parentRect = null;
+    var parentComputedStyle = window.getComputedStyle(parent);
```
CaptainCodeman: This broke things for me on Firefox and Safari (I think it's when the parent is the html element). The version before this change works fine.

philipwalton: Can you go into a bit more detail here? What specifically broke? `getComputedStyle()` should work on the `<html>` element, and if the `<html>` element is set to `display: none`, I think exiting early is still the correct logic.

CaptainCodeman: Yes, sorry - that wasn't a lot to go on.

I'm using it within WebComponents and the above change made everything stop working with an error on that line, I'll have to update again to get the error but the gist of it was that `getComputedStyle` was being called on something that didn't support it, I thought it was the document root.

I'll find the error and raise a proper issue ...

Issue #234, opened 2017-08-02
Polyfill: Chrome uses Polyfill instead of native code


jesusbotella:
Hello,

I have just started using Intersection Observer to improve the performance of my web, and I noticed that if I import the polyfill in Chrome it is always used, instead of the native code.

The reason why it is always used is because of the condition used to determine if it should be loaded or not. It checks if the property [`intersectionRatio` exists in `IntersectionObserverEntry` prototype](https://github.com/WICG/IntersectionObserver/blob/gh-pages/polyfill/intersection-observer.js#L25), and it is evaluated to false.

What I see in [Intersection Observer Entry docs on MDN](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry) is that the intersectionRatio property is available from Chrome 51, and I am currently using Chrome 60. So I don't know why the value is false.

Should the condition be changed in order not to load the polyfill in Chrome?

Thank you!

philipwalton:
What makes you say it's evaluated to false? The condition evaluates to true for me in Chrome and the native version is used.

jesusbotella:
Yes, I'm sorry, the condition evaluates to true in a chrome blank window, but I'm bundling the polyfill inside another package. I think that due to JS hoisting and the IntersectionObserverEntry definition below, when I try to evaluate `'intersectionRatio' in IntersectionObserverEntry.prototype` just before evaluating the if condition, it is evaluated to false.

However, if you check if there exists `'intersectionRatio' in window.IntersectionObserverEntry.prototype`, it is true because you are checking it with the global definition of the IntersectionObserverEntry. I tried to modify it, and it works as expected.

I don't know if I'm missing something, but I think that it could be the reason why it was failing. 

If you think I'm correct, I'd be happy with making a PR to fix the code, or feel free to do it yourself if better.

samouss:
I agree with @jesusbotella, [the recent changes](https://github.com/WICG/IntersectionObserver/commit/d4a699cc7a2d12e4e67cd6d1e748bed8bb6e3d8c#diff-d881855d8287e00bed364a1cee1dd6b6) on the polyfill update the condition.

philipwalton:
@nolanlawson, sounds like this is related to the change in your PR (#224). Can you PTAL? I'm OOO for the next two weeks so I won't have time to dig into it.

pottedmeat:
[...]

PR #235, opened 2017-08-17
Stop functions being hoisted and interfering with the feature detect …

JakeChampion:
…by wrapping functions in an IIFE

This should be a solution for #234

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -37,689 +37,690 @@ if ('IntersectionObserver' in window &&
   return;
 }
 
-
-/**
- * An IntersectionObserver registry. This registry exists to hold a strong
- * reference to IntersectionObserver instances currently observering a target
- * element. Without this registry, instances without another reference may be
- * garbage collected.
- */
-var registry = [];
-
-
-/**
- * Creates the global IntersectionObserverEntry constructor.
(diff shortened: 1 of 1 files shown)


nolanlawson:
This LGTM; although I don't have the privileges to merge this PR. @philipwalton?

BTW GitHub doesn't do a great job with the diff, but essentially you're adding an extra

```js
(function(window, document) {
 /* ... */
}(window, document));
```

after the `return`. Perhaps a simpler solution, which would avoid the extra IIFE, is to remove the `return` in favor of a big `if`/`else` instead? (I'm assuming that module bundlers are causing weirdness due to the `return`.)

philipwalton:
I've merged #239 as I think it's a more explicit solution to #234 and it has the benefit of not introducing another closure.

astorije:
> BTW GitHub doesn't do a great job with the diff

You may already be aware of it, but in case you or anyone stops by and is not, GitHub has a very nifty trick for this: add [`?w=1`](https://github.com/w3c/IntersectionObserver/pull/235/files?w=1) at the end of the URL, and it will hide all whitespace-only changes (effectively equivalent to `git diff -w` in the terminal), showing this PR as 3 lines of changes only 😁 

Only caveat that I know of is that it will also completely hide inline comments in PRs, which can be deceiving.

Hope that helps :)

PR #239, opened 2017-08-19
Look for intersectionRatio on the global IntersectionObserverEntry

### LINKED ISSUE #234: Polyfill: Chrome uses Polyfill instead of native code

pottedmeat:
Currently looks at the local IntersectionObserverEntry's prototype when doing feature detection rather than the global (native) constructor function.

Fixes #234

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -22,7 +22,7 @@
 // features are natively supported.
 if ('IntersectionObserver' in window &&
     'IntersectionObserverEntry' in window &&
-    'intersectionRatio' in IntersectionObserverEntry.prototype) {
+    'intersectionRatio' in window.IntersectionObserverEntry.prototype) {
 
   // Minimal polyfill for Edge 15's lack of `isIntersecting`
   // See: https://github.com/WICG/IntersectionObserver/issues/211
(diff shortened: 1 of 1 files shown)


samouss:
I think you should add the `window` prefix on the next condition and the minimal polyfill for Edge, because the same hoisting problem will occur.

https://github.com/pottedmeat/IntersectionObserver/blob/381f8f09d1d24746a2508da6a428115f8662a845/polyfill/intersection-observer.js#L29

samouss:
Cool! But the call of `Object.defineProperty` suffer of the same problem.

https://github.com/pottedmeat/IntersectionObserver/blob/381f8f09d1d24746a2508da6a428115f8662a845/polyfill/intersection-observer.js#L30

philipwalton:
Thanks!

PR #240, opened 2017-08-19
Check for intersections each time a node is observed

### LINKED ISSUE #238: Calling .observe after monitoring has begun does not result in an initial entry event
The .observe function does not call `_checkForIntersections` which means that adding observe to a node that was already in DOM will not receive an initial entry and we will not know its status until one of the other events triggers an intersection check.

It is illustrated in this demo: https://jsfiddle.net/xoLk8ppp/

pottedmeat:
If a node is being observed that was already in DOM and monitoring has already begun, an intersection check must be performed to the throttled `_checkForIntersections`.

Fixes #238

### DIFF
--- a/polyfill/intersection-observer-test.js
+++ b/polyfill/intersection-observer-test.js
@@ -178,6 +178,20 @@ describe('IntersectionObserver', function() {
       io.observe(targetEl2);
     });
 
+    it('triggers for existing targets when observing begins after monitoring has begun', function(done) {
+      var spy = sinon.spy();
+      io = new IntersectionObserver(spy, {root: rootEl});
+
+      io.observe(targetEl1);
+      setTimeout(function() {
+        io.observe(targetEl2);
+        setTimeout(function() {
+          expect(spy.callCount).to.be(2);
+          done();
+        }, ASYNC_TIMEOUT);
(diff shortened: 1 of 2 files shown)


philipwalton:
If you look at the [implementation of `_monitorIntersections()`](https://github.com/w3c/IntersectionObserver/blob/bf4f257a5160c59fc6a81dad5fc397e7a10268a1/polyfill/intersection-observer.js#L259), you'll see that it calls `_checkForIntersections()` if it's not already monitoring, so I think there must be something else going on to create your issue.

pottedmeat:
Yes, you can see how it happens in the linked issue. Like you said, if it's already monitoring, it fails to check for intersections.

philipwalton:
Ahh, I think I misunderstood you at first. Now I see what you're saying.

Can you update your PR to include a test that checks for this? If there's not test that's currently failing in this case, there should be. If you don't have time, I'll add add one when I get a chance, but I don't want to merge this without a test.

pottedmeat:
Test has been added. I verified that this test fails without the fix.

philipwalton:
Thanks! Last thing is `this._checkForIntersections()` probably doesn't need to happen in `_monitorIntersections()` anymore. Other than that, LGTM.

PR #245, opened 2017-08-28
Fire intersection events on isIntersecting changes

### LINKED ISSUE #222: Incompatibilities between spec and implementations for observer events
(Forking discussion from https://github.com/WICG/IntersectionObserver/issues/211)

There are incompatibilities between Chrome, Edge, Firefox, and the spec on when exactly observer events should fire.

According to the spec:

> Let thresholdIndex be the index of the first entry in observer.thresholds whose value is greater than intersectionRatio, or the length of observer.thresholds if intersectionRatio is greater than the last entry in observer.thresholds.

And

> If thresholdIndex does not equal previousThresholdIndex, queue an IntersectionObserverEntry

So IntersectionObserverEntrys (events) should only be fired when the `thresholdIndex` changes. However, what we've observed in our testing is that Chrome fires this event when `isIntersecting` changes, regardless of whether `thresholdIndex` changes (which we can observe through `intersectionRatio`).

So for example, [here is a test](http://bl.ocks.org/nolanlawson/raw/5b23c154cd231af8b252677a2507fcdc/) showing Chrome firing an extra event for an element whose `intersectionRatio` does _not_ change but whose `isIntersecting` has changed. And here are the cross-browser results:

out

And [here is a test](http://bl.ocks.org/nolanlawson/raw/eb6aab0a76126a4e08545813ecad585e/) showing Chrome _not_ firing an event for an element whose `isIntersecting` hasn't changed, but whose `intersectionRatio` has changed. And here are the cross-browser results:

out2

So roughly, here's the results with a ✔️ for event fired, and ❌ for event not being fired:

| | Spec | Chrome 59.0.3071.109 | Edge 15.15063 | Firefox Developer Edition 55.0b3 |
| --- | --- | --- | --- | --- |
| Demo 1 | ❌ | ✔️ |  ❌ | ✔️ |
| Demo 2 | ✔️ | ❌ | ✔️ | ✔️ |

Currently we (EdgeHTML) are weighing the options in our own implementation, but I'm interested to hear what Blink/Gecko folks think about this.

nolanlawson:
Fixes #222. This change should match Blink's existing behavior and EdgeHTML 16's behavior. See https://github.com/w3c/IntersectionObserver/issues/222#issuecomment-311189229 and https://github.com/w3c/IntersectionObserver/issues/222#issuecomment-325415567 for justification.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -229,7 +229,7 @@ interface IntersectionObserver {
 			an {{IntersectionObserverRegistration}} record
 			with an {{IntersectionObserverRegistration/observer}} property set to |this|,
 			a {{IntersectionObserverRegistration/previousThresholdIndex}} property set to <code>-1</code>,
-			and a {{IntersectionObserverRegistration/previousIsIntersecting}} property set to <code>undefined</code>.
+			and a {{IntersectionObserverRegistration/previousIsIntersecting}} property set to <code>false</code>.
 		3. Append |intersectionObserverRegistration|
 			to |target|'s internal {{[[RegisteredIntersectionObservers]]}} slot.
 		4. Add |target| to |this|'s internal {{[[ObservationTargets]]}} slot.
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -501,9 +502,9 @@ Element</h4>
 which is initialized to an empty list.
 This list holds <dfn interface>IntersectionObserverRegistration</dfn> records,
 which have an <dfn attribute for=IntersectionObserverRegistration>observer</dfn> property
-holding an {{IntersectionObserver}}
-and a <dfn attribute for=IntersectionObserverRegistration>previousThresholdIndex</dfn> property
-holding a number between -1 and the length of the observer's {{IntersectionObserver/thresholds}} property (inclusive).
+holding an {{IntersectionObserver}}, a <dfn attribute for=IntersectionObserverRegistration>previousThresholdIndex</dfn> property
+holding a number between -1 and the length of the observer's {{IntersectionObserver/thresholds}} property (inclusive), and
+a <dfn attribute for=IntersectionObserverRegistration>previousIsIntersecting</dfn> property holding a boolean or <code>undefined</code>.
```
szager-chromium: Can we get rid of the "undefined" value for previousIsIntersecting, and just initialize it to "false"?  The -1 value for previousThresholdIndex will still serve the purpose of forcing the first notification, so it seems like we can safely make previousIsIntersecting and strict boolean.

nolanlawson: Good catch! Yes, we can keep it strict boolean.

PR #247, opened 2017-08-29
Add intersectionRatio and isIntersecting to the prototype

JakeChampion:
Adding these to the prototype will make sure that basic feature detection will pass for the polyfilled version of IntersectionObservers.

Running this on a browser which has IntersectionObserver natively will return true but on a browser with the polyfill will currently return false.
```
'IntersectionObserver' in this &&
'IntersectionObserverEntry' in this &&
'intersectionRatio' in this.IntersectionObserverEntry.prototype &&
'isIntersecting' in this.IntersectionObserverEntry.prototype
```

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -59,23 +59,38 @@ function IntersectionObserverEntry(entry) {
   this.rootBounds = entry.rootBounds;
   this.boundingClientRect = entry.boundingClientRect;
   this.intersectionRect = entry.intersectionRect || getEmptyRect();
-  this.isIntersecting = !!entry.intersectionRect;
-
-  // Calculates the intersection ratio.
-  var targetRect = this.boundingClientRect;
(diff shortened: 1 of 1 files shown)


philipwalton:
Can you explain what you're trying to accomplish a bit more here? What is the use case for having polyfills pass feature detection (I'm not sure I'd want that)?

Also, doesn't this still fail depending on how the feature detection is implemented?

JakeChampion:
Thinking on this more, the introduction of defineProperty will make this not work in IE7/8. Closing this PR.

What I was aiming to accomplish was to get the polyfill closer to how native versions are implementated. I know this is not completely possible due to native versions not allowing IntersectionObserverEntry from being constructed directly and that Object.prototype.toString.call(window.IntersectionObserver) will return '[native code]' for native versions but, I figured I'd have a go at getting the polyfill be able to pass it's own feature detection script is a good thing. Most polyfills can pass their own feature detection on https://polyfill.io.

PR #249, opened 2017-08-29
Use rAF with a timeout of 0 and test

### LINKED ISSUE #248: Use requestAnimationFrame when THROTTLE_TIMEOUT is 0
To match the native API as closely as possible, calculations should be performed in a `requestAnimationFrame` callback. Because calculations are expensive when done in a polyfill, `requestAnimationFrame` should only be used when `THROTTLE_TIMEOUT` is 0.

pottedmeat:
Use `requestAnimationFrame` instead of `setTimeout` when `THROTTLE_TIMEOUT` is 0 and the API is available. Added test.

Resolves #248.

### DIFF
--- a/polyfill/intersection-observer-test.js
+++ b/polyfill/intersection-observer-test.js
@@ -805,6 +805,35 @@ describe('IntersectionObserver', function() {
       io.observe(targetEl1);
     });
 
+
+    it('uses requestAnimationFrame when timeout is 0', function(done) {
+      var rAF = window.requestAnimationFrame;
+      if (!rAF) {
+        done();
+      }
+      else {
+        var timeout = IntersectionObserver.prototype.THROTTLE_TIMEOUT;
+        var stub = sinon.stub(window, 'requestAnimationFrame').returns(1);
+
+        // without a THROTTLE_TIMEOUT of 0, setTimeout should be used
+        IntersectionObserver.prototype.THROTTLE_TIMEOUT = timeout || 100;
+        io = new IntersectionObserver(function() {}, {root: rootEl});
+        io.observe(targetEl1);
+        expect(stub.callCount).to.be(0);
+        io.disconnect();
+
+        // when THROTTLE_TIMEOUT is 0 and requestAnimationFrame is available, it should be used
+        IntersectionObserver.prototype.THROTTLE_TIMEOUT = 0;
+        io = new IntersectionObserver(function() {}, {root: rootEl});
+        io.observe(targetEl1);
+        expect(stub.callCount).to.be(1);
+
+        window.requestAnimationFrame = rAF;
+        IntersectionObserver.prototype.THROTTLE_TIMEOUT = timeout;
+        done();
+      }
+    });
+
   });
 
 
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -552,7 +552,7 @@ function now() {
 
 
 /**
- * Throttles a function and delays its executiong, so it's only called at most
+ * Throttles a function and delays its execution, so it's only called at most
  * once within a given time period.
  * @param {Function} fn The function to throttle.
  * @param {number} timeout The amount of time that must pass before the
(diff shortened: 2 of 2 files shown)


philipwalton:
Hmmm, but couldn't someone still use a timeout of 1ms (or even `'foo'`, `NaN`, `null`, `undefined`, or anything else that's effectively a timeout of 0)?

If the concern is making it impossible to footgun by having a too-short timeout, then we should cover all of these cases. On the other hand, if what you want is the ability to use `requestAnimationFrame` as a throttle, then maybe we could add a configuration option to support that.

What do you think?

pottedmeat:
The idea of the fix is meant to be pretty simple - `requestAnimationFrame` doesn't support an arbitrary timeout so it should only be used for a timeout of 0. Whereas `setTimeout` supports an arbitrary timeout but doesn't make the same guarantees that `requestAnimationFrame` does - and the native implementation expects callbacks to happen in the next animation frame.

jdreetz:
Won't this prevent the callback from firing when it's in a background tab? Does that match the native IntersectionObserver behavior? Wouldn't that prevent the callback from firing with "out of view" when the user switches tabs? Haven't tested it, so I could be wrong.

philipwalton:
I discussed with spec editor @szager-chromium and it looks like [the spec changed a while back](https://github.com/w3c/IntersectionObserver/commit/2bad9ae2c173978b6be4ef754a10bb1f32b2f191#diff-ec9cfa5f3f35ec1f84feb2e59686c34d) to remove the 100ms timeout that the original polyfill code was based on. Now a more spec compliant version would be to keep the `throttle` method using `setTimeout`, but to set the default timeout to `0`.

Since all the work the polyfill does to detect intersections is quite expensive, I'd be in favor of keeping the existing behavior that allows authors to set their own throttle period, but updating the default to `0`.

What do others think?

siusin:
Please merge the requested change to the `master` branch (`index.bs`) once the group/editors reach a consensus on this issue, the travis bot will generate an` index.html` with the latest version of Bikeshed and push it to [`gh-pages`](https://w3c.github.io/IntersectionObserver/).

Our apologies for the inconvenience.

pottedmeat:
Rebased onto master. Is still reporting as conflicted.

siusin:
Thanks @pottedmeat . Could you try [update the base-branch](https://help.github.com/articles/changing-the-base-branch-of-a-pull-request/) of your pull request?

pottedmeat:
Done. Thanks.

szager-chromium:
I don't have any real objection to the patch, but I think there may be some general confusion about the timing of things...

In a spec-compliant (i.e. native) implementation, the "check intersections" step is scheduled separately from the "invoke callbacks" step.  "Check intersections" happens immediately *after* a frame is generated.  Conceptually, it's like a setTimeout that magically pre-empts any other pending timers.  So a frame looks like this:

```
1. Run pending timers (expired setTimeout's and setInterval's)
2. Run requestAnimationFrame callbacks
3. Generate a frame (layout/paint/raster)
4. Check intersections
5. Go to step 1
```

During the "check intersections" step, if new notifications are generated, they are scheduled using a setTimeout(0).  That means that they're scheduled along with any other pending setTimeout's and setInterval's:

```
Frame 1:
  - Run pending timers (setTimeout's, setInterval's)
  - Run requestAnimationFrame callbacks
  - Generate a frame
  - Check intersections, schedule notifications
Frame 2:
  - Run pending timers, including notifications generated in Frame 1
  - rAF
  - Generate a frame
  - Check intersections
...
```

When the polyfill runs the "check intersections" code, it delivers notifications synchronously.  It may be OK for the "check intersections" code to be scheduled during rAF, although this will introduce a somewhat arbitrary delay in the calculation.  Probably not a big deal.  It may, however, be surprising for the user-code callbacks to run during rAF, rather than during "pending timers."

I'm a very inexperienced web developer, so I will yield to the consensus, but I want to make sure everyone understands this and has considered the effect of this change.

pottedmeat:
This may not be the best approach based on the feedback. I'll do a bit more research and follow up.

miketaylr:
Given the lack of activity on this PR, I propose that we close it. If someone would like to pick the work back up, we can re-open. Thanks.

PR #252, opened 2017-09-04
Change links that point at WICG to point at w3c

robinwhittleton:
Now that the spec has moved from https://github.com/WICG to https://github.com/w3c there are a few parts of the code and documentation that need updating. Some redirect automatically, but others currently 404.


siusin:
Merged.

Thanks @robinwhittleton !

Issue #261, opened 2017-10-20
MutationObserver on Safari 10.0.3 causes browser to freeze and CPU usage to spike to 100%


bpossolo:
The MutationObserver activated by [this block of code](https://github.com/w3c/IntersectionObserver/blob/master/polyfill/intersection-observer.js#L270) causes my Safari browser (10.0.3) to grind to a standstill.

I would like to submit a PR to make the usage of the MutationObserver optional.
It would look something like this:

```
IntersectionObserver.prototype.USE_MUTATION_OBSERVER = true;
...
if (this.USE_MUTATION_OBSERVER && 'MutationObserver' in window) {
 // existing code that registers MutationObserver
}
```

usage:
```
var io = new IntersectionObserver(...);
io.USE_MUTATION_OBSERVER = false;
```

Any feedback is welcome.

philipwalton:
I'm not opposed to something like this since I'm sure there are sites (e.g. content sites) where the DOM rarely changes and thus MO use isn't really needed, but I'm also curious to know why this is happening for you.

I've not previously heard that MO is slow on Safari, so I'm wondering if this is specific to your use case.

JSteunou:
This issue can now be closed :)

Thank you everyone involved!

dackmin:
I know this issue was closed a while back, but I am still experiencing this with Safari 11.1, even with `USE_MUTATION_OBSERVER` set to `false` 🤷‍♂️  

@bpossolo @JSteunou do you remember doing anything else back then not to make Safari and/or your CPU suffocate when using the polyfill ? Beside not using it 😄

PR #265, opened 2017-10-27
Make mutation observer optional & better confined

JSteunou:
2 changes: 

1. Make mutation observer optional (solves #261) because it is rarely needed
2. Use `root` options to prevent observing the all document when only a tiny part is scrollable. Prevent triggering unnecessary check on intersection when somewhere the document is highly dynamic.

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -131,6 +131,12 @@ IntersectionObserver.prototype.THROTTLE_TIMEOUT = 100;
  */
 IntersectionObserver.prototype.POLL_INTERVAL = null;
 
+/**
+ * Use a mutation observer on the root element 
+ * to detect intersection changes.
+ */
+IntersectionObserver.prototype.USE_MUTATION_OBSERVER = true;
+
 
 /**
  * Starts observing a target element for intersection changes based on
@@ -267,9 +273,9 @@ IntersectionObserver.prototype._monitorIntersections = function() {
       addEvent(window, 'resize', this._checkForIntersections, true);
       addEvent(document, 'scroll', this._checkForIntersections, true);
 
-      if ('MutationObserver' in window) {
+      if (this.USE_MUTATION_OBSERVER && 'MutationObserver' in window) {
         this._domObserver = new MutationObserver(this._checkForIntersections);
-        this._domObserver.observe(document, {
+        this._domObserver.observe(this.root || document, {
           attributes: true,
           childList: true,
           characterData: true,


JSteunou:
Can someone review this PR? @siusin or @philipwalton ?

JSteunou:
Someone ? Anyone ? friendly ping @philipwalton @restrry @marcoscaceres ?

polyfill/intersection-observer.js
```diff
@@ -131,6 +131,12 @@ IntersectionObserver.prototype.THROTTLE_TIMEOUT = 100;
  */
 IntersectionObserver.prototype.POLL_INTERVAL = null;
 
+/**
+ * Use a mutation observer on the root element 
+ * to detect intersection changes.
+ */
+IntersectionObserver.prototype.USE_MUTATION_OBSERVER = true;
```
marcoscaceres: I don't think this is part of the spec, so shouldn't be here. Otherwise, you should use a ES6 Symbol, right?

polyfill/intersection-observer.js
```diff
@@ -267,9 +273,9 @@ IntersectionObserver.prototype._monitorIntersections = function() {
       addEvent(window, 'resize', this._checkForIntersections, true);
       addEvent(document, 'scroll', this._checkForIntersections, true);
 
-      if ('MutationObserver' in window) {
+      if (this.USE_MUTATION_OBSERVER && 'MutationObserver' in window) {
```
marcoscaceres: As above, this shouldn't be publicly exposed if not in the spec, otherwise people using the polyfill might end up doing the wrong thing.

polyfill/intersection-observer.js
marcoscaceres: Also, it's unclear if user land code should be able to set this? Should it be read only.

JSteunou: I'm not using Symbol because this is a polyfill and should be compatible with a wide range of user agent. Also this is the way this polyfill already expose flags https://github.com/w3c/IntersectionObserver/blob/master/polyfill/intersection-observer.js#L117 

Also this PR addresses #261 which @philipwalton seems to be ok with

polyfill/intersection-observer.js
JSteunou: This is the point of this flag, to be publicly exposed in order to deactivate mutation observer listeners which are very resource consuming and not needed in a lot of cases. (I would not say 99% but meh :p)

marcoscaceres: You still risk authors becoming reliant on this non-standard behavior - which you definitely don't want in a polyfill. 

cc @triblondon for guidance.

bpossolo: This polyfill is an implementation of the IntersectionObserver spec.
The use of the MutationObserver is an implementation detail and has nothing to do with the spec.
What @JSteunou has submitted is simply a way to configure this particular implementation which any user of the polyfill should be permitted to do.

I don't see how there is any risk in a user becoming reliant on this configuration property any more than they are already reliant on the polyfill to begin with.

polyfill/intersection-observer.js
bpossolo: @marcoscaceres the mutationobserver makes the polyfill unusable on certain browsers and consumes more resources than are often necessary so it definitely should be configurable by users.

I agree with @JSteunou that this is exactly like the other flags and is implemented correctly.

marcoscaceres: > @marcoscaceres the mutationobserver makes the polyfill unusable on certain browsers and consumes more resources than are often necessary so it definitely should be configurable by users.

I'm completely sympathetic to what you are saying @bpossolo, but if there are clear perf wins from having when not using mutation observers, then always take the middle ground. If the polyfill starts sprouting additional attributes, etc. then it's no longer a polyfill, as it no longer matches the spec, which can come back to bite us:

https://www.w3.org/2001/tag/doc/polyfills/

Cc @bkardell

polyfill/intersection-observer.js
marcoscaceres: But it looks like it's modifying the interface definition on the global scope:

```JS
IntersectionObserver.prototype.USE_MUTATION_OBSERVER = true;
```

What am I missing? If my understanding is correct, then you've added something non-standard to the API.

bpossolo: the configuration properties are orthogonal to the usage of the IntersectionObserver spec api.
the polyfill already provides 2 configuration properties and the [README](https://github.com/w3c/IntersectionObserver/tree/master/polyfill#configuring-the-polyfill) describes how to use them.

here's this new config property might be used:

my polyfills.ts
```
import 'intersection-observer/intersection-observer';
IntersectionObserver.prototype.USE_MUTATION_OBSERVER = false;
// if the browser natively supports IntersectionObserver, then the polyfill
// is a no-op and the property is moot
```

my app.ts
```
const options = {
  threshold: 0.25
};
const onIntersection = function(entries: IntersectionObserverEntry[], observer: IntersectionObserver): void {
  // do stuff with intersection entries
};
const observer = new IntersectionObserver(onIntersect, options);
// do stuff with observer
```

JSteunou: @marcoscaceres if you're against exposing configuration flag on `IntersectionObserver` prototype then you should remove `THROTTLE_TIMEOUT` and `POLL_INTERVAL` for consistency.

marcoscaceres: Yeah. Just thinking out loud here, but maybe put `POLL_INTERVAL` and friends into the container scope or as an `configOptions` on the bootstrap:
```js 
(function(window, document, configOptions) {
   const { THROTTLE_TIMEOUT,  POLL_INTERVAL, USE_MUTATION_OBSERVER } = configOptions; 
   // other stuff
}(window, document, {
  THROTTLE_TIMEOUT: false, 
  POLL_INTERVAL: false,  
  USE_MUTATION_OBSERVER: false,
}));
 ```

Comment (JSteunou) [reply polyfill/intersection-observer.js]: How do you expect users to configure use of this polyfill this way? oO

Comment (marcoscaceres) [reply polyfill/intersection-observer.js]: They could change the values manually, which is admittedly not great.  Another option would be to set it inline:

```HTML
<script>
var interOps = {
  THROTTLE_TIMEOUT: false, 
  POLL_INTERVAL: false,  
  USE_MUTATION_OBSERVER: true,
}
</script>
<script src="https://the.poly.fill/intersection.js"></script> 
```

And then the polyfill would have:

```JS
{
const defaultOps = {
  THROTTLE_TIMEOUT: false, 
  POLL_INTERVAL: false,  
  USE_MUTATION_OBSERVER: false,
};
(function(window, document, configOptions) {
  const { THROTTLE_TIMEOUT,  POLL_INTERVAL, USE_MUTATION_OBSERVER } = {
     // blend together 
     ... defaultOption, ...configOptions
  }; 
   // other stuff
}(window, document,  interOps ||  defaultOps));
```

Or some such. There are probably lots of ways of doing it - that's just one off the top of my head. 

Another might be using the URL of the script:

```HTML
<script src="https://the.poly.fill/intersection.js?THROTTLE_TIMEOUT=true"></script> 
```

JSteunou: This is not as simple from a developer and user of this polyfill point of view. I'm speaking for my case but there are a lot of different usage & context. I'm using polyfill service and build with webpack, I do not add global variable neither query parameters on my build upon my app's views needs and usage, and I'm pretty sure everyone would have an example in which your changes would make them miserable.

Why one would make it difficult for everyone rather than keep it simple? I think the point of a polyfill is to help spread the use of new feature. If not, it would be never use until a decade.

In French we have a say I like a lot that can be like this in English 

> perfectionism can be counter-productive 

Meanwhile, why just not validate this PR, which solve an issue for users and you would make another one with your changes about those flag, moving this debate into the right place without blocking this one.

polyfill/intersection-observer.js
triblondon: fwiw, I'm relaxed about polyfills exposing non-spec attributes if their implementation incorporates trade offs that developers might want access to / have an opinion on.  It's also sometimes useful to make the polyfill detectable.  So the principle of perfect spec conformance is not hard and fast for me (or for the TAG polyfill recommendations).  In this specific case, I think I suggested using MOs in the first place, to avoid polling.  Polyfill.io tries to avoid shipping polyfills that poll, because they make it hard to debug applications when your flamechart is basically just noise.

polyfill/intersection-observer.js
bpossolo: Refactoring the configuration properties seems outside of the scope of this PR and is derailing.

polyfill/intersection-observer.js
bpossolo: The way I view it...the spec is simply an interface and the implementations of any interface will naturally differ and should be configurable.

This implementation chooses to use a MutationObserver and polling.
The native browser implementation likely uses something completely different and will likely expose their own native properties on their objects.

JSteunou:
Thank you @marcoscaceres 

Could this PR be marked as *[non-substantive](https://labs.w3.org/hatchery/repo-manager/pr/id/w3c/IntersectionObserver/265)* and merged?

marcoscaceres:

polyfill/intersection-observer.js
```diff
@@ -267,9 +273,9 @@ IntersectionObserver.prototype._monitorIntersections = function() {
...
         this._domObserver = new MutationObserver(this._checkForIntersections);
-        this._domObserver.observe(document, {
+        this._domObserver.observe(this.root || document, {
```
philipwalton: Sorry for the late comment (I was on vacation), but I'm not sure this should have been added. There are cases where mutations to an element outside of the root element could affect the positioning of one of the observed elements. So if we want to be safe, we should listen to all mutations.

JSteunou: I understand the issue here. At first I though about a kind of resize sensor on the root element, but like you said, not only size but also positioning could be affected.
As users can now use the `USE_MUTATION_OBSERVER` to cut off unrelated spaming events to the watched element this one can be rolled back.

PR #267, opened 2017-11-10
Delete LICENSE

marcoscaceres:
Things in this repo should use the w3c license


philipwalton:
This PR is against the gh-pages branch, but it looks like master is the base branch for the repo, and it already doesn't include the LICENSE file.

I'm not sure what's needed for the spec to be displayed, but we probably shouldn't have both of these branches...

marcoscaceres:
@philipwalton, thanks! didn't see that. I've filed a bug. We good if we can pick one, otherwise it's confusing and things could fall out of sync.

PR #273, opened 2017-12-04
Add more doc about Mutation Observer

JSteunou:


JSteunou:
@philipwalton I though adding some doc about the previous PR can be better than no doc, especially with the all document being listened by default.

polyfill/README.md
```diff
@@ -81,6 +81,18 @@ io.observe(someTargetElement);
 
 **Note:** the `POLL_INTERVAL` property must be set prior to calling the `.observe` method, or the default configuration will be used.
 
+**Dom changes through MutationObserver**
```
philipwalton: s/Dom/DOM/

I'd remove the mention of `MutationObserver` in the title here since it's primarily an implementation detail (after thinking about it a bit more, arguably the name could change as well).

Maybe change it to something like "Ignoring DOM changes".

polyfill/README.md
```diff
@@ -81,6 +81,18 @@ io.observe(someTargetElement);
...
+
+You can also choose to not listen to the changes to the DOM with `USE_MUTATION_OBSERVER`. [MutationObserver](https://developer.mozilla.org/fr/docs/Web/API/MutationObserver) is used when available to listen changes to the DOM that could affect the elements intersections. But it can come with a high cost if your website / webapp contains high frequency updated elements unrelated to the intersections watched elements. If you are absolutely sure you do not need to listen to DOM changes, you can deactivate it globally or for an instance
```
philipwalton: I'd suggest rewording this to something like:

You can also choose to not check for intersections when the DOM changes by setting an observer's `USE_MUTATION_OBSERVER` property to `false` (either globally on the prototype or per-instance).

This is recommended in cases where the DOM will update frequently but you know those updates will have no affect on the position or your target elements.

JSteunou:
Here you go ;)

philipwalton:

PR #277, opened 2017-12-14
Prevent setting the polyfill multiple times on further loads

klipstein:
When the IntersectionObserver polyfill is loaded multiple times it should not be reset.

Due to this check https://github.com/w3c/IntersectionObserver/blob/master/polyfill/intersection-observer.js#L18 the polyfill will be overwritten on every load.

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -69,6 +69,10 @@ function IntersectionObserverEntry(entry) {
   }
 }
 
+/**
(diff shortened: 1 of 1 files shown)


siusin:

miketaylr:
> When the IntersectionObserver polyfill is loaded multiple times

Thanks for the PR, but loading the script multiple times feels like a bug that should be addressed by site operators

PR #279, opened 2018-01-22
Server Side Render friendly version

Ornhoj:
Wrapped entire IIFE in an if block, avoiding checking undefined window object.

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -6,719 +6,720 @@
  *  https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
  *
  */
-
-(function(window, document) {
-'use strict';
-
-
-// Exits early if all IntersectionObserver and IntersectionObserverEntry
-// features are natively supported.
-if ('IntersectionObserver' in window &&
-    'IntersectionObserverEntry' in window &&
-    'intersectionRatio' in window.IntersectionObserverEntry.prototype) {
-
-  // Minimal polyfill for Edge 15's lack of `isIntersecting`
-  // See: https://github.com/w3c/IntersectionObserver/issues/211
-  if (!('isIntersecting' in window.IntersectionObserverEntry.prototype)) {
-    Object.defineProperty(window.IntersectionObserverEntry.prototype,
-      'isIntersecting', {
-      get: function () {
-        return this.intersectionRatio > 0;
-      }
-    });
-  }
-  return;
-}
+if (typeof window !== undefined) {
+    (function(window, document) {
+    'use strict';
 
 
-/**
- * An IntersectionObserver registry. This registry exists to hold a strong
- * reference to IntersectionObserver instances currently observering a target
- * element. Without this registry, instances without another reference may be
- * garbage collected.
- */
-var registry = [];
+    // Exits early if all IntersectionObserver and IntersectionObserverEntry
+    // features are natively supported.
+    if ('IntersectionObserver' in window &&
+        'IntersectionObserverEntry' in window &&
+        'intersectionRatio' in window.IntersectionObserverEntry.prototype) {
 
-
-/**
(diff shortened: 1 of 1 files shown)


philipwalton:
Can you give a little more context here? Why is this needed? Does the server-side rendering process execute this code?

Ornhoj:
@philipwalton Thank you for following up.

When JS is serverside rendered, all imports are taken into consideration, and executed if the code does not prevent this upfront. The results is that any interaction with the window object will throw a `window is undefined` and break further execution of the code.

By wrapping the entire IIFE in an if condition, we would avoid executing any of the IIFE code as long as the JS is executed server side where there are no window object available.

**Summery:**
I believe it would be best practice avoiding execution of the code in a content where there are no window available - like my case with serverside rendering.

nadnad:
Running into this problem when using this package on a site built with Gatsby. Can confirm that this PR fixes the issue.

The Gatsby error:
```
  723 |     window.IntersectionObserverEntry = IntersectionObserverEntry;
  724 |
> 725 |   }(window, document));
      |    ^
  726 | }
  727 |
  728 |


  WebpackError: window is not defined

  - intersection-observer.js:725 Object.exports.__esModule
    ~/intersection-observer/intersection-observer.js:725:4
```

chrstntdd:
Although having this polyfill contain the logic to play well on the server, I've been able to achieve the expected behavior by adding the logic you've introduced before requiring the polyfill. Is there any drawback to handling it on a case by case basis?

```js
if (typeof window !== 'undefined') {
  require('intersection-observer');
}
```

Ornhoj:
@chrstntdd handling this on a case to case basis is one way to do it - I prefer both as an author and a user to making a more robust code. Hence making sure that server side rendering is not hindered should be a baseline, since it has been getting a lot of tracktion

philipwalton:
Rather than wrap the entire library in an `if` statement, would it also work to replace references to `window` with `self`?

Ornhoj:
@philipwalton technically yes - however the issue at hand is that the entire polyfill is an IIFE with the window and document objects as arguments.
Hence avoiding execution if either of these is unavailable would also be a minor performance boost.

If the wrapping is unacceptable an inline appoach might be posible By testing the passed window and document objects at the very beginning of the IIFE e.g:
`(function(_window, _document) {
    var isBrowser = (typeof _window !== 'undefined') && (typeof _document !== 'undefined');

})(window, document)`

And in every place where the passed window or document objects are referenced we would prefix the request with `isBrowser &&` .

This setup is a lot more verbose and in my stylistic opinion simply wrong. Why enter the function when the fundamentals of the function is undefined and causes errors?

TechQuery:
@philipwalton 
Test scripts based on **JSDOM** & **ES 6+** (Babel) will execute `import` before the fake `window` & `document` are ready. ([Example](https://github.com/EasyWebApp/WebCell/blob/master/test/DOM-polyfill.js))

JulianJorgensen:
Hey guys, whats the status on the merge? Would really love to give this a try!

philipwalton:

philipwalton:
Sorry all for the slow response. @Ornhoj I've marked your PR as non-substantive, so there's no longer an IP issue and this can be merged once there are no merge conflicts. Can you rebase against master?

philipwalton:
Also, can you use a check similar to what's suggested in this comment? https://github.com/w3c/IntersectionObserver/pull/339#issuecomment-486437268

philipwalton:
I've manually merged this. It should be available in version `0.7.0`.

koundinyab:
@philipwalton Can you please dumb down the solution to make the intersectionObserver work with angular universal please ?

philipwalton:
@koundinyab what about it currently doesn't work? Also, it's probably better if you file a new issue since this PR is already merged.

PR #290, opened 2018-04-02
Judgement target's nodeType at first

guox191:
Hello,

I think it's more logical to validate the target's type at first, then checks if it is already observed.

Thanks.


siusin:

PR #294, opened 2018-04-19
AnimationEnd and TransitionEnd Observation also added

kashifeqbal:


szager-chromium:
Polyfill is no longer supported, closing.

Issue #291, opened 2018-04-11
The intersection observation algorithm doesn't match reality


mrbkap:
While investigating a [related bug in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1391154), I started looking into edge cases in the intersection observation code and found that both Firefox and Chrome's existing behavior differ from what is specified.

The problem has to do when the smallest threshold value is greater than 0. Consider [this testcase](https://gist.github.com/mrbkap/3ac333fca1659be5c60932c259cec8e2). We're observing div `d1`. It starts not intersecting, all implementations agree that the first output should be "record[0]: false (0)". Then, as it moves to the right, the right edge of the div clips the left edge of the root element. At this point, the algorithm in the spec would have a `previousThresholdIndex` of 0 and `previousIsIntersecting` of `false`. Now that `d1` intersects the root, `isIntersecting` would be true, causing us to queue an `IntersectionObserverEntry`. However, neither Firefox nor Chrome do this. They both treat all `intersectionRatio`s smaller than the smallest value in `observer.thresholds` as equivalent.

So, my expected output would be:
```
record[0]: false (0)
record[0]: true (0.5)
```

I believe that the algorithm in the spec (assuming I'm not mistaken) would give something like:

```
record[0]: false (0)
record[0]: true (0)
record[0]: true (0.5)
```

This makes intuitive sense to me (not having studied any use cases) as I would think the smallest threshold value would be the smallest `intersectionRatio` that my observer would be notified about. If I cared about the edge where `isIntersecting` was changing, I'd make sure that I had a threshold value of 0.

I believe we should update the algorithm in the spec to match the current browsers' implementations.

One additional nitpick: step 9 in the algorithm specifies ```Let thresholdIndex be the index of the first entry in observer.thresholds whose value is greater than intersectionRatio, or the length of observer.thresholds if intersectionRatio is greater than or equal to the last entry in observer.thresholds.``` It isn't immediately clear what to do in the case where `intersectionRatio` is smaller than all values in `observer.thresholds`. I believe it would clarify matters to explicitly specify ```if no entry in observer.thresholds is greater than intersectionRatio, let thresholdIndex be 0```.

szager-chromium:
Please take a look at the pull request and let me know if it clarifies things.

PR #316, opened 2018-09-07
Fix edge case for reporting transitions

szager-chromium:
If the first threshold value is non-zero, then no notification should
happen if isIntersecting changes but thresholdIndex does not.

Issue #291


emilio (commented):
I need to go through this one a bit more closely, but I'm not sure this section of the spec generally matches browsers, see https://github.com/w3c/IntersectionObserver/issues/432.

szager-chromium:
> I need to go through this one a bit more closely, but I'm not sure this section of the spec generally matches browsers, see #432.

Yeah, it would be a good idea to incorporate that issue into this patch. I will also dig into it.

szager-chromium:
I added a comment to issue #432 with, I hope, a thorough analysis.

Issue #274, opened 2017-12-07
"Set this.root to options.root." doesn't really make sense


smaug----:
"Set this.root to options.root." doens't really have meaning.
The interface has attribute root, but that is just a getter method.

szager-chromium:
Please take a look at the pull request, and let me know if it adequately addresses this.

PR #318, opened 2018-09-07
Fix specification of IntersectionObserver.root attribute in constructor

szager-chromium:
Issue #274

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/pull/318.html" title="Last updated on Sep 2, 2020, 7:18 PM UTC (eaea2bf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/318/7e57e90...eaea2bf.html" title="Last updated on Sep 2, 2020, 7:18 PM UTC (eaea2bf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -212,20 +212,20 @@ interface IntersectionObserver {
 	::
 		1. Let |this| be a new {{IntersectionObserver}} object
 		2. Set |this|'s internal {{[[callback]]}} slot to |callback|.
-		3. Set |this|.|root| to |options|.{{IntersectionObserverInit/root}}.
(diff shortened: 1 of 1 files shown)


emilio (commented):
Probably the specification for the root getter should say something more concrete than:

> The root Node to use for intersection, or null if the observer uses the implicit root.

Probably just saying that this is the root node specified in the options passed to the constructor would be enough. A more formal definition may use an internal slot or something.

emilio (approved):

emilio:
(I probably missed the last line of the diff? huh)

Issue #297, opened 2018-05-02
Does the polyfill support iframes?


canscaw28:
I would like use the polyfill within an iframe to detect if the iframe itself has been intersected with the host's viewport. Is this supported? Thanks!

szager-chromium:
If the iframe is cross-origin, then no, it will not work. I *think* same-origin iframes are supported, but not 100% sure.

richardraviis:
I am also facing issue with the cross origin iframe. Can someone please confirm if the polyfill works on cross origin iframe.

philipwalton:
The polyfill does not currently work on either same-origin or cross-origin iframes.

While it *could* work on same-origin iframes, supporting that would add a significant amount of extra code to the size of the polyfill, and I originally I wasn't sure how common of a use case it would be.

The amount and complexity of code required to support iframes is large because you have to know the x/y coordinates of the current frame you're in with respect to the top window. While that in and of itself isn't too bad, where it gets really tricky is handling cases where the frame is nested within one or more parent frames and those frames may be moving/scrolling/transitioning their positions, thus affecting your frame's x/y position inside the top frame's window.

My sense is most iframe use cases are cross-origin anyway, so unless I hear a lot of people asking for this feature, I'd prefer to keep the implementation minimal.

dackmin:
@philipwalton you should at least mention it in the polyfill README, as the current browser implementation supports cross-origin iframes correctly (at least on Firefox & Chrome).
I spent a lot of time wondering why it wasn't working on Safari until I noticed Safari was using the polyfill.

philipwalton:
I added a [limitations](https://github.com/w3c/IntersectionObserver/tree/master/polyfill#limitations) section where I address this and the proposed v2 changes (which the polyfill will also not be able to support).

PR #321, opened 2018-09-24
allow IntersectionObserver polyfill to get loaded in an iframe 

gevou:
This fix is possibly related to issue #297 

When the polyfill is loaded inside an <iframe> with the `root` option set to `null` then the root viewport is  the <iframe> itself and therefore does not track any activity on the parent/host window. This was observed on both mobile and desktop versions of Safari. 
However, this is not the default behavior for `IntersectionObserver` on browsers with native support e.g. Chrome, in which case the `root` viewport is correctly calculated with respect to the top window.

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -9,24 +9,27 @@
 
 (function(window, document) {
 'use strict';
-
+var _loadedWindowOrIFrame = window;
+window = window.top;
+document = window.document;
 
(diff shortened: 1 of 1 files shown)


dontcallmedom:

philipwalton:
These changes don't account for the relative position of an element within a frame with respect to the top window, and supporting that use case is quite a bit more complicated.

I go into more detail here on why I don't think we need to support this: https://github.com/w3c/IntersectionObserver/issues/297#issuecomment-431679486

PR #323, opened 2018-10-08
Use the defined root as scrolling container when setted.

AlineLee:


szager-chromium:
Closing this as the polyfill is no longer supported.

Issue #324, opened 2018-10-11
Intersection computation error in IE11


ShakMR:
Given how the operation between floats in javascript works in IE11 computing the intersection with the observed element can result in `0.99999581...` and, given a threshold of 1, the callback is not called.

Could a solution be rounding the intersecting ratio to some decimal (for example 3rd)? Would this fit into the way we want the intersection to be computed?

philipwalton:
Do you have an example page showing an element fully visible but the ratio < 1? I understand that this is technically possible in JS, I'm just trying to see how common it is in reality.

Also, if we do rounding, I'd want to be sure we avoid a situation where we round to the wrong value and make the problem worse.

As for how many decimal points to round to, I'd say if we go this route it should be at least 4 so two-thirds is represented as `66.67%` vs. `66.7%`.

ShakMR:
I don't have access to IE at the moment to check if it's already there in some part of the site (I'll check tomorrow). Anyway, I couldn't reproduce this on simple test (with just some DIVs and SPANs). 

Regarding the decimals, 4 more makes sense, yes.

>Also, if we do rounding, I'd want to be sure we avoid a situation where we round to the wrong value and make the problem worse.
Make sense, in any other case I'll be OK with the behaviour, but in this case I imagine than using 1 as a threshold could be a normal usage and a rounding problem that gives you 0.999995 is quite unexpected. It took me almost a week to find it out.

Also, I don't know which precision other use cases may need. I terms of rounding been a problem.

amn:
I strongly disagree -- I've seen these floating point errors crop up, and I believe the error lies in the browsers and how they calculate bounds. Patching `IntersectionObserver` to follow erroneous user agent behavior seems like a major step in the wrong direction.

If anything, adding such rounding up and down of values should be at least implemented as an extra feature per observer object and not something that cannot be opted out of otherwise.

philipwalton:
Before taking any action here, I'd like to see a repro where the polyfill reports a meaningfully different value from the native implementation. I'm going to close this issue for now, but if anyone can provide that I'd gladly look into updating the polyfill to more closely reflect the native behavior.

PR #325, opened 2018-10-16
Rounding polyfill intersection ratio to 3 decimals

ShakMR:
In the intersection observer draft, it talks about threshold as a ratio using examples of 1 decimal only. 
I thought of 3 decimals as I normally think on the intersection ratio as a percentage of the area so we can match thresholds up to 3 decimals (or percentages up to 1 decimal).

https://github.com/w3c/IntersectionObserver/issues/324

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -62,7 +62,7 @@ function IntersectionObserverEntry(entry) {
 
   // Sets intersection ratio.
   if (targetArea) {
-    this.intersectionRatio = (intersectionArea / targetArea).toFixed(3) - 0;
+    this.intersectionRatio = Number((intersectionArea / targetArea).toFixed(4));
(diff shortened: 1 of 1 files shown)


dontcallmedom:

polyfill/intersection-observer.js
```diff
@@ -62,7 +62,7 @@ function IntersectionObserverEntry(entry) {
 
   // Sets intersection ratio.
   if (targetArea) {
-    this.intersectionRatio = intersectionArea / targetArea;
+    this.intersectionRatio = (intersectionArea / targetArea).toFixed(3) - 0;
```
philipwalton: Can you use either `Number()` or `parseFloat()` here instead of `- 0` to do the type conversion? I understand your way is shorter, but it's less clear why it's happening.

PR #331, opened 2018-11-30
Fix the timing of running the IntersectionObserver algorithm

szager-chromium:
As discussed in the TAG review, IntersectionObserver should run its
algorithm *after* requestAnimationFrame callbacks, not before.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -657,8 +657,7 @@ HTML Processing Model: Event Loop</h4>
 
 An <a>Intersection Observer</a> processing step should take place
 during the "<i>Update the rendering</i>" steps,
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@szager-chromium, thanks! Is there a web platform test to verify the new behavior?

szager-chromium:
@marcoscaceres All of the existing WPT tests already expect that behavior; the previous spec language was just wrong, and I'm pretty sure all of the browsers just ignored it and did the Right Thing.

marcoscaceres:
Excellent! thank you for confirming.

PR #339, opened 2019-01-16
 Modify IIFE to safely execute in Node environments

kohlmannj:
## Description
This PR modifies the polyfill's IIFE wrapper to safely allow execution within a Node environment without `window` and `document` globals. The modified IIFE uses a combination of these techniques:

- [IIFE for non-browser global environments](https://toddmotto.com/what-function-window-document-undefined-iife-really-means/#non-browser-global-environments)
- Pass either `window` or `this` to the wrapped function, [as seen in jQuery's wrapper](https://github.com/jquery/jquery/blob/e743cbd28553267f955f71ea7248377915613fd9/src/wrapper.js#L41)

## Motivation + Background
I've been working on an internal project which uses this polyfill to support Safari on macOS and iOS for our usages of a React `<Observer>` component ([@kohlmannj/react-intersection-observer](
https://github.com/kohlmannj/react-intersection-observer)).

I started by statically importing the polyfill (e.g. `import 'intersection-observer';`), but found that our webpack and Node-based pre-render process (without JSDOM, so without `window` and `document` globals) exited with an error due to the polyfill's IIFE wrapper.

I then tried a few variations of the theme of dynamically `import()`ing of the polyfill client-side, and only in browsers that needed it. While this worked, it added significant complexity to our usage of the `<Observer>` component (since applying the polyfill is a global side effect, I identified some tricky React render order concerns).

Therefore, I finally went back to this polyfill to see if I could address the root cause of the Node pre-render's failure. I then researched and combined the two generally accepted approaches mentioned above to create the modified IIFE.

Finally, I noted that the test suite in intersection-observer-test.html passed all tests when run in Safari 12.0.2 (14606.3.4) on macOS Mojave 10.14.2. This suggests that the modifications have no impact on the polyfill's correctness when applied client-side.

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -729,4 +729,4 @@ function getParentNode(node) {
 root.IntersectionObserver = IntersectionObserver;
 root.IntersectionObserverEntry = IntersectionObserverEntry;
 
-}(this));
+}(typeof window !== "undefined" ? window : this));
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -7,20 +7,20 @@
  *
  */
 
-(function(window, document) {
+(function(root) {
 'use strict';
 
 
 // Exits early if all IntersectionObserver and IntersectionObserverEntry
 // features are natively supported.
-if ('IntersectionObserver' in window &&
-    'IntersectionObserverEntry' in window &&
-    'intersectionRatio' in window.IntersectionObserverEntry.prototype) {
+if ('IntersectionObserver' in root &&
+    'IntersectionObserverEntry' in root &&
+    'intersectionRatio' in root.IntersectionObserverEntry.prototype) {
 
   // Minimal polyfill for Edge 15's lack of `isIntersecting`
   // See: https://github.com/w3c/IntersectionObserver/issues/211
-  if (!('isIntersecting' in window.IntersectionObserverEntry.prototype)) {
-    Object.defineProperty(window.IntersectionObserverEntry.prototype,
+  if (!('isIntersecting' in root.IntersectionObserverEntry.prototype)) {
+    Object.defineProperty(root.IntersectionObserverEntry.prototype,
       'isIntersecting', {
       get: function () {
         return this.intersectionRatio > 0;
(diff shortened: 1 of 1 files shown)


kohlmannj:
FYI, I have linked my GitHub account with a newly created W3C account! 🎉

dontcallmedom:

kohlmannj:
@dontcallmedom Thank you! Excited to see this merged soon as well 😅😄

winsandymyint:
Could you please merge this version to master?
We're getting face an error due to the polyfill's IIFE wrapper.

philipwalton:
This is a duplicate of #279. @kohlmannj in your opinion is there any reason to prefer one over the other?

polyfill/intersection-observer.js
```diff
@@ -726,7 +726,7 @@ function getParentNode(node) {
 
 
 // Exposes the constructors globally.
-window.IntersectionObserver = IntersectionObserver;
-window.IntersectionObserverEntry = IntersectionObserverEntry;
+root.IntersectionObserver = IntersectionObserver;
+root.IntersectionObserverEntry = IntersectionObserverEntry;
 
-}(window, document));
+}(typeof window !== "undefined" ? window : this));
```
kohlmannj: In the future, this line could be modified to support `globalThis`:

```suggestion
}((
  typeof globalThis !== "undefined"
    ? globalThis
    : typeof window !== "undefined"
       ? window
        : this
));
```

**Note:** I would prefer a version which does _not_ use nested ternary statements, so this is for illustration purposes only.

kohlmannj:
@philipwalton Hi! A few things, I guess:

1. Compared to #279, this PR is a much smaller diff
2. The approach from this PR follows a well-defined IIFE convention (as seen in [jQuery](https://github.com/jquery/jquery/blob/e743cbd28553267f955f71ea7248377915613fd9/src/wrapper.js#L41))
3. The refactor of the IIFE contents from `window` to `root` is beneficial since this code no longer assumes the name of a particular global object
4. Related, the change needed to support [the `globalThis` proposal](https://github.com/tc39/proposal-global) in the future will be one line: https://github.com/w3c/IntersectionObserver/pull/339/files#r278318268

philipwalton:
From my perspective, the primary purpose of both of these PRs it to allow the code to be imported/required in a non-window environment and not error in the initial evaluation.

If imported in a non-window environment, I don't think it should be adding symbols to the global object if it's not actually going to be used.

My preferred solution would be one that lets it evaluate successfully in any JavaScript environment but only run (i.e. modify global state) in a window environment.

kohlmannj:
@philipwalton Ah, I see what you mean now — it's "execute safely" (this PR) vs. "don't execute" (#279).

In that case, I'm in favor of closing this PR in favor of #279. I would prefer a slightly more comprehensive check in #279, however, such as this check from [tuxsudo/is-in-browser](https://github.com/tuxsudo/is-in-browser/blob/master/src/index.js#L1-L3):

```js
const isBrowser = typeof window === "object"
    && typeof document === 'object'
    && document.nodeType === 9;
```

Thank you for clarifying 👍🏼

philipwalton:
A fix for this should be in `0.7.0`.

PR #340, opened 2019-01-16
Add `IntersectionObserver.prototype.IS_POLYFILL`

kohlmannj:
This PR adds property `IS_POLYFILL` to the IntersectionObserver prototype. This allows code that runs after the polyfill to detect if the client-side IntersectionObserver implementation is non-native, e.g. in cases where the developer wants to disable certain behavior for performance reasons.

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -132,6 +132,12 @@ IntersectionObserver.prototype.POLL_INTERVAL = null;
  */
 IntersectionObserver.prototype.USE_MUTATION_OBSERVER = true;
 
+/**
+ * Indicates that IntersectionObserver has been polyfilled, allowing
+ * other code to detect the presence of the polyfilled implementation.
+ */
+IntersectionObserver.prototype.IS_POLYFILL = true;
+
 
 /**
  * Starts observing a target element for intersection changes based on


kohlmannj:
FYI, I have linked my GitHub account with a newly created W3C account! 🎉

dontcallmedom:

kohlmannj:
@dontcallmedom Thank you! Excited to see this merged soon.

mcgoooo:
@kohlmannj  thanks for this, using this myself now!

JakeChampion:
There is no need for this because you can detect any polyfills via the Function.prototype.toString method like so: `Function.prototype.toString.call(IntersectionObserver) === "function IntersectionObserver() { [native code] }"`

nickion:
> There is no need for this because you can detect any polyfills via the Function.prototype.toString method like so: `Function.prototype.toString.call(IntersectionObserver) === "function IntersectionObserver() { [native code] }"`

You have just illustrated why there is a need for this.

JakeChampion:
`Function.prototype.toString.call(IntersectionObserver) === "function IntersectionObserver() { [native code] }"` returns false for the polyfill and true for the native implemenation

kohlmannj:
> `Function.prototype.toString.call(IntersectionObserver) === "function IntersectionObserver() { [native code] }"` returns false for the polyfill and true for the native implementation.

If that resulting string is consistent cross-browser and will remain consistent for the foreseeable future, I can live with that! Good thinking, thanks @JakeChampion.

philipwalton:
Yes, @JakeChampion is correct. Checking for "[native code]" works cross-browser and is a common way user-code can detect native implementations vs. polyfills.

The only time this doesn't work is when polyfills modify the `toString()` method to return "[native code]" themselves, but that doesn't apply here since this polyfill doesn't do that.

PR #350, opened 2019-02-25
Added AMD version 

verlok:
Hey there,
I need an AMD version of the script in order to conditionally load it _only there it's needed_ using Require.JS, so I made it!

If you accept this, maybe others could benefit from it too.
Otherwise it will live only in my little, humble [repository](https://github.com/verlok/IntersectionObserverAMD).

### DIFF
--- a/polyfill/intersection-observer-amd.js
+++ b/polyfill/intersection-observer-amd.js
@@ -1,0 +1,736 @@
+define(function () { 'use strict';
+
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the W3C SOFTWARE AND DOCUMENT NOTICE AND LICENSE.
+ *
+ *  https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
+ *
+ */
+
+(function(window, document) {
+    'use strict';
+    
+    
+    // Exits early if all IntersectionObserver and IntersectionObserverEntry
+    // features are natively supported.
+    if ('IntersectionObserver' in window &&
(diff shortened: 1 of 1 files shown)


philipwalton:
My understanding is RequireJS can load plain JavaScript files, and since this polyfill doesn't have any dependencies, I don't believe it needs a `define()` wrapper. Am I wrong?

verlok:
I didn’t know that RequireJS could load plain JavaScript files as a dependency. 

So you think that something like this could work without requiring an AMD version of the polyfill?

verlok:
Found this in the RequireJS doc 

> Ideally the scripts you load will be modules that are defined by calling define(). However, you may need to use some traditional/legacy "browser globals" scripts that do not express their dependencies via define(). For those, you can use the shim config. To properly express their dependencies.

> If you do not express the dependencies, you will likely get loading errors since RequireJS loads scripts asynchronously and out of order for speed.

Now I’m trying to understand what the shim config is.

verlok:
Found the shim config doc here:
https://requirejs.org/docs/api.html#config-shim

This means that in order to load this polyfill in AMD, the user would need to play and fiddle with the RequireJS config file. Which can be tricky, but doable. 

You decide what to do here. I prefer not to touch the RequireJS config, and I can keep loading my own version of the polyfill in AMD.

philipwalton:
All browsers currently support native JavaScript modules, and from what I see most AMD usage is gradually switching to native JavaScript modules anyway.

My preference would be to not support older module formats if those formats have a way to solve this on their own.

PR #363, opened 2019-05-23
Provide a minified version through unpkg.com CDN

WebReflection:
Using the field "unpkg" in a generic `package.json` file is a common
technique to enable the delivery of a module through the unpkg.com CDN.

As example, the following script will automatically bring in
the minified version of the current polyfill.
```html
<script src="//unpkg.com/intersection-observer"></script>
```

This MR would like to provide both the `unpkg` field and a minified version of the polyfill, so that it'll be easier for anyone to bring it in, without unnecessary bloat.


szager-chromium:
Closing this as the polyfill is no longer supported.

PR #369, opened 2019-06-16
Fix possible exception in old Firefox versions

dudabone:
In older versions of Firefox, window.getComputedStyle may return null when element is invisible (see https://bugzilla.mozilla.org/show_bug.cgi?id=548397). This causes an exception in the polyfill

Closes #???

Implementation commitment:


szager-chromium:
Closing this, as the polyfill is no longer supported.

PR #386, opened 2019-09-18
Define content clip

szager-chromium:
Issue#351

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/pull/386.html" title="Last updated on Feb 3, 2021, 4:25 AM UTC (e89f811)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/386/7e57e90...e89f811.html" title="Last updated on Feb 3, 2021, 4:25 AM UTC (e89f811)">Diff</a>


index.bs
```diff
@@ -304,6 +308,8 @@ or else the <a>top-level browsing context</a>'s <a>document</a> node
 (referred to as the <dfn for="IntersectionObserver">implicit root</dfn>) if
 the {{IntersectionObserver/root}} attribute is <code>null</code>.
 
+An {{Element}} is defined as having an <dfn for="IntersectionObserver">overflow clip</dfn> if its computed style has a value other than "visible" for its <a>overflow</a>, <a>overflow-x</a>, or <a>overflow-y</a> property.
```
emilio, line 311: "overflow clip" seems quite an odd term because `clip` is a valid value of the `overflow` property, fwiw.

There's no need to mention the overflow property explicitly, as it's just a shorthand of `overflow-x` and `overflow-y`.

index.bs
```diff
@@ -581,7 +587,7 @@ run these steps:
 2. Let |container| be the <a>containing block</a> of the <a>target</a>.
 3. While |container| is not the <a>intersection root</a>:
 	1. Map |intersectionRect| to the coordinate space of |container|.
-	2. If |container| has overflow clipping or a css <a>clip-path</a> property,
+	2. If |container| has an <a>overflow clip</a> or a CSS <A>clip-path</a> property,
```
emilio, line 590: The casing change at the left of `clip-path` seems unintentional.

emilio (changes requested):
I think this should probably use https://drafts.csswg.org/css-overflow/#scroll-container instead.

There are elements for which the `overflow-x/y` styles don't necessarily create a scroll container (e.g., an `` element, or an element with `display: none`).

szager-chromium:
> I think this should probably use https://drafts.csswg.org/css-overflow/#scroll-container instead.
> 
> There are elements for which the `overflow-x/y` styles don't necessarily create a scroll container (e.g., an `` element, or an element with `display: none`).

"Scroll container" is not quite right, because it is defined as an element that can be scrolled by the user, whereas IntersectionObserver should be applying the clip regardless of whether the element is user scrollable. The first part of that spec section refers to CSS properties that "specify whether a box’s content (including any ink overflow) is clipped to its padding edge". I've updated the language to specify that IntersectionObserver will apply any padding edge clip mandated by that spec section.

index.bs
szager-chromium: I changed this to "content clip".

index.bs
szager-chromium: Fixed.

emilio:
> it is defined as an element that can be scrolled by the user

Hmm, how so? `<div style="overflow: hidden"></div>` is not scrollable by the user but is a scroll container, right? I think the definition there is a bit poor but https://drafts.csswg.org/css-overflow/#valdef-overflow-hidden makes it clear...

szager-chromium:
> > it is defined as an element that can be scrolled by the user
> 
> Hmm, how so? `<div style="overflow: hidden"></div>` is not scrollable by the user but is a scroll container, right? I think the definition there is a bit poor but https://drafts.csswg.org/css-overflow/#valdef-overflow-hidden makes it clear...

Here's the first line in that section of the spec:

"These properties specify whether a box’s content (including any ink overflow) is clipped to its padding edge, and if so, whether it is a **_scroll container_** that allows the user to scroll clipped parts of its scrollable overflow area into view."

So, "scroll container" is a subset of elements that have their content clipped to the padding edge; specifically, it refers to those elements that are user-scrollable. For IntersectionObserver, we want to apply the padding edge clip to the superset, including overflow:hidden and overflow:clip.

emilio:
I think that's a bug in the css spec fwiw. `<div style="overflow: hidden">` totally is a scroll container. But I `overflow: clip` isn't so... I guess this is fine to account for both.

index.bs
```diff
@@ -305,6 +307,8 @@ or else the <a>top-level browsing context</a>'s {{document}} node
 (referred to as the <dfn for="IntersectionObserver">implicit root</dfn>) if
 the {{IntersectionObserver/root}} attribute is <code>null</code>.
 
+An {{Element}} is defined as having a <dfn for="IntersectionObserver">content clip</dfn> if its computed style has <a>overflow properties</a> that cause its content to be clipped to the element's padding edge.
```
emilio, line 310: Hmm, this makes me think... this doesn't deal with the `clip` property at all, and it maybe should?

emilio (approved):

PR #389, opened 2019-09-18
Allow an empty string for rootMargin.

szager-chromium:
Issue#330

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/pull/389.html" title="Last updated on Sep 2, 2020, 5:39 PM UTC (eb78616)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/389/989e687...eb78616.html" title="Last updated on Sep 2, 2020, 5:39 PM UTC (eb78616)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -349,21 +349,23 @@ or failure:
 1. <a>Parse a list of component values</a> |marginString|,
 	storing the result as |tokens|.
 2. Remove all whitespace tokens from |tokens|.
-3. If the length of |tokens| is 0 or greater than 4,
+3. If the length of |tokens| is greater than 4,
 	return failure.
-4. Replace each |token| in |tokens|:
+4. If there are zero elements in |tokens|,
+	set |tokens| to ["0px"].
+5. Replace each |token| in |tokens|:
 	* If |token| is an <a>absolute length</a> <a>dimension</a> token,
 		replace it with a an equivalent pixel length.
 	* If |token| is a <<percentage>> token,
 		replace it with an equivalent percentage.
 	* Otherwise, return failure.
-5. If there is one element in |tokens|,
+6. If there is one element in |tokens|,
 	append three duplicates of that element to |tokens|.
-	If there is two elements are |tokens|,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -345,19 +345,21 @@ or failure:
...
 	return failure.
 4. Replace each |token| in |tokens|:
 	* If |token| is an <a>absolute length</a> <a>dimension</a> token,
 		replace it with a an equivalent pixel length.
 	* If |token| is a <<percentage>> token,
 		replace it with an equivalent percentage.
 	* Otherwise, return failure.
-5. If there is one element in |tokens|,
+5. If there are zero elements in |tokens|,
+	set |tokens| to ["0", "0", "0", "0"].
```
emilio, line 357: This should be done before parsing, right? Otherwise you probably need to word it as "an absolute lengths of zero pixels" or something.

Also, this doesn't match Blink, as far as I can tell. Per this change, `new IntersectionObserver(() => {}, { rootMargin: " " })` should end up with a zero margin, while it seems like Blink special-cases the empty string.

emilio (changes requested):
It's a bit sad that instead of actually implementing the spec WebKit just copied Blink...

Can we have a WPT test for this? I'm happy to update Firefox to match other browsers here...

index.bs
emilio: Before parsing also doesn't work, because 0 is a number token, not a dimension or percentage token.

szager-chromium: I changed it to apply the default value before parsing. I also added WPT coverage, and fixed chromium's behavior:

https://chromium-review.googlesource.com/c/chromium/src/+/2388706/1/third_party/blink/web_tests/external/wpt/intersection-observer/observer-attributes.html

emilio (approved):

emilio:
Please remove the merge commit before merging though.

szager-chromium:
> Please remove the merge commit before merging though.

I will do it, under protest :)

Issue #372, opened 2019-07-02
Cannot track intersection with an iframe's viewport


szager-chromium:
The implicit root refers to the top window's scrolling viewport, but there is currently no way to track the intersection of an element inside an iframe with the iframe window's scrolling viewport. We might try something like this (running in the iframe's context):

new IntersectionObserver(callback, {root: document.scrollingElement})

... but that doesn't work, because document.scrollingElement.getBoundingClientRect gives you the full size of the document's content, not the window's scrolling viewport.

I think document.scrollingElement ought to be treated specially when it's used as the root of an IntersectionObserver: it should be treated as measuring intersection with the document window's scrolling viewport.

chrishtr:
AMP documents would benefit from this. In many cases, content within AMP is rendered within an iframe. The lack of support for this use-case means AMP can't use IntersectionObserver for such situations.

chrishtr:
The proposal as stated would be a change of semantics without any IDL change, meaning it's technically not backwards-compatible. However, @szager-chromium and I agree that there seems to be no good use-case for supplying an explicit root that is a scrolling element without such a clip, so we expect usage of the API in this mode to be almost non-existent.

chrishtr:
Also, there is an easy workaround if the other behavior is desired: make sure the <html> element is the scrolling element, and observe relative to <body>.

emilio:
Copying from an email thread:

Off-hand the idea seems fine, I agree this is a really unfortunate limitation. It seems unfortunate to change the semantics of the root option though... Maybe we should extend to allow `root` take a window or a document instead, and use that to represent the viewport root?

That'd make `{root: window}` or `{root: document}` on a top level document behave as no-root, and on an iframe behave as you want with respect to the viewport.

I haven't thought it through all that much though, but given the `scrollingElement` can change over time as a result of style changes, at least in quirks mode documents, if you define it as you propose you need to decide whether the "is scrolling element" check happens at the moment you create the observer (which would mean that creating an observer would need to update layout, which is bad), or at the moment the notifications are sent, or at some other point (when?).

It seems it'd cause confusion / subtle bugs either way though.

emilio:
cc @mstange / @dholbert in case they have other opinions.

szager-chromium:
FYI, we landed this change behind a flag in chromium:

https://chromium-review.googlesource.com/c/chromium/src/+/1865257

In this implementation, the (root == document.scrollingElement) check happens every time we run the intersection algorithm.

[...]

Issue #95, opened 2016-02-08
Use visual viewport instead of layout viewport for root intersection rectangle?


ChumpChief:
The definition of the root intersection rectangle specifies that for intersection roots which are document viewports, it is specifically the layout viewport.

I think this should instead be the visual viewport to better align with the stated goal in the spec of "calculating element visibility".

There is a corresponding note that this choice was intentional, implying there was intent of keeping the root intersection rectangle closely related to layout dimensions (specifying that pinch zoom changes neither).  If that is still true and this change is rejected, then the relationship between IntersectionObserver and layout dimensions should be better explained.  The goals outlined in the spec currently sound more relevant to "what the user sees" than "how the page is laid out".

I would also be concerned that leaving this as the layout viewport would leave the "exotic plugin-based solutions" for ad impression tracking as more-effective than IntersectionObserver and discourage uptake.

Alternatively, if there are scenarios which require intersection against the layout viewport, then an option to specify layout vs. visual viewport as the root intersection rectangle could be interesting.

szager-chromium:
You are correct that the intention is visual viewport; marking this "needs spec".

ChumpChief:
Sounds good, thanks.  As a followup, it should probably be made explicit what the coordinate space is for rootMargin then.

I think it should be measured in the visual viewport's coordinate space, e.g. such that 100vh represents content that is one page away, or approximately one "page down" press away.  As opposed to the layout viewport's coordinate space, where 100vh in that coordinate space may be multiple "page down" presses away depending on the user's pinch zoom level.

szager-chromium:
Actually, I'm sorry, my previous reply was incorrect.  After clarifying this with some of the people involved:

At least for the first release, IntersectionObserver will track intersections with the layout viewport, not the visual viewport.  The primary reason for this is that by tracking the layout viewport, we are not revealing any new information to cross-origin iframes than they currently have.  IntersectionObserver will provide the same information that is currently available, but it will do it more accurately and more efficiently (in terms of performance and battery consumption) by eliminating the need for pollers or scroll event handlers.

Tracking intersections with the visual viewport, while unquestionably useful, is also likely to be more controversial, so we're punting on that for now, pending more discussion with browser vendors.

ojanvafai:
@ChumpChief that sound OK to you for now?

ChumpChief:
[...]

PR #411, opened 2020-01-21
Allow the 'root' argument to be a Document.

szager-chromium:
Issue #372

Closes #???

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/szager-chromium/IntersectionObserver/pull/411.html" title="Last updated on Jan 23, 2020, 8:01 PM UTC (6515fa7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/411/d3eece4...szager-chromium:6515fa7.html" title="Last updated on Jan 23, 2020, 8:01 PM UTC (6515fa7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -168,11 +168,11 @@ The IntersectionObserver interface</h3>
 The {{IntersectionObserver}} interface can be used to observe changes in the
 intersection of an <a>intersection root</a> and one or more <a>target</a> {{Element}}s.
 
-An {{IntersectionObserver}} with a {{IntersectionObserver/root}} {{Element}} can
+An {{IntersectionObserver}} with a {{IntersectionObserver/root}} {{Node}} can
 observe any <a>target</a> {{Element}} that is a descendant of the
 {{IntersectionObserver/root}} in the <a>containing block chain</a>.
 
-An {{IntersectionObserver}} with no {{IntersectionObserver/root}} {{Element}}
+An {{IntersectionObserver}} with no {{IntersectionObserver/root}} {{Node}}
 will automatically observe intersections with the <a>implicit root</a>,
 and valid <a>targets</a> include any {{Element}} in the
 <a>top-level browsing context</a>, as well as any {{Element}} in any
@@ -195,7 +195,7 @@ introduces a need for it.
 [Constructor(IntersectionObserverCallback callback, optional IntersectionObserverInit options),
  Exposed=Window]
 interface IntersectionObserver {
-	readonly attribute Element? root;
+	readonly attribute Node? root;
 	readonly attribute DOMString rootMargin;
 	readonly attribute FrozenArray&lt;double&gt; thresholds;
 	void observe(Element target);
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -310,9 +310,12 @@ is the rectangle we'll use to check against the targets.
 
 <dl class=switch>
 	<dt>If the <a>intersection root</a> is the <a>implicit root</a>,
-	<dd>it's the viewport's size.
+	<dd>it's treated as if the root were the <a>top-level browsing context</a>'s {{document}}, according to the following rule for {{document}}.
 
-	<dt>If the <a>intersection root</a> has an overflow clip,
+	<dt>If the <a>intersection root</a> is a {{document}},
+	<dd>it's the size of the {{document}}'s viewport.
```
bzbarsky: Where is "the document's viewport" defined?  Is it defined for documents that are not in a browsing context or are not fully active?

szager-chromium: This is sort of intentionally vague. The existing spec says this:

"If the intersection root is the implicit root, it [the root intersection rectangle] is the viewport’s size."

Elsewhere it mentions "the viewport of the document" and "viewport of the nested browsing context". It also says this in a highlighted note:

"Root intersection rectangle is not affected by pinch zoom and will report the unadjusted viewport, consistent with the intent of pinch zooming (to act like a magnifying glass and NOT change layout.)"

However, all of that text is likely to change with the resolution of this issue:

https://github.com/w3c/IntersectionObserver/issues/95

The intention here is to use "viewport" in a way that is consistent with existing usage in the spec, until the above issue is resolved.

bzbarsky: I understand all that, but the implicit root, being the active document of a toplevel browsing context, always has at least _some_ concept of a viewport.  If there's some random document involved (e.g. an XHR responseXML document, or a document from an iframe that is now removed from the DOM or whatever), there may be no concept of a viewport at all.  #95 does not cover that, as far as I can tell.

szager-chromium: The only spec language I could find that defines browser viewport is this unhelpful bit:

https://drafts.csswg.org/css2/visuren.html#viewport

But I'm not sure it's necessary to determine what the browser viewport of a detached document is, since the "Update the rendering" steps in the HTML event loop -- which is where IntersectionObserver runs -- only applies to fully active (connected) documents:

https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model

bzbarsky: Aha.  It might be good in the loop to assert that all documents we encounter are fully active.  Those might still not have a sane viewport (e.g. if their embedding iframe is display:none), but at least it cuts down on the cases that need to be considered.

szager-chromium: OK, I added some language there to clarify that the root document must be fully active at this step in processing.

PTAL

szager-chromium:
@bzbarsky If you could take another look at this soon-ish, that would be helpful. We're hoping to ship this behavior in the next branch of chromium, which will happen next week.

Thanks

index.bs
bzbarsky: This looks reasonable to me, thank you.

szager-chromium: @bzbarsky Would you mind clicking the "approved" button...

PR #419, opened 2020-04-08
Polyfill: support same-origin iframe targets

dvoytenko:
The polyfill is updated to support observing targets in the same-origin iframes. This is already a spec'd behavior and only polyfill abilities are enhanced. See https://www.w3.org/TR/intersection-observer/#dom-intersectionobserver-rootmargin for more info.

Notably, the changes to do not polyfill the `IntersectionObserver` inside an iframe, but simply allow observing using the top-level context `IntersectionObserver`. The final interface binding can be done by manually:

```
iframe.contentWindow.IntersectionObserver = window.IntersectionObserver;
iframe.contentWindow.IntersectionObserverEntry = window.IntersectionObserverEntry;
```

In practice, however, I find that even when same-origin iframes are used, the root observer is used for them. The tests show this pattern.

### DIFF
--- a/polyfill/intersection-observer-test.html
+++ b/polyfill/intersection-observer-test.html
@@ -27,6 +27,14 @@ Licensed under the W3C SOFTWARE AND DOCUMENT NOTICE AND LICENSE.
   <!-- Setup -->
   <script>mocha.setup({ui:'bdd'});</script>
 
+  <!-- Force polyfill. -->
+  <script>
+    if (false) {
+      delete window.IntersectionObserver;
+      delete window.IntersectionObserverEntry;
+    }
+  </script>
+
   <!-- Script under test -->
   <script src="intersection-observer.js"></script>
 
--- a/polyfill/intersection-observer-test.html
+++ b/polyfill/intersection-observer-test.html
@@ -27,13 +27,12 @@ Licensed under the W3C SOFTWARE AND DOCUMENT NOTICE AND LICENSE.
   <!-- Setup -->
   <script>mocha.setup({ui:'bdd'});</script>
 
-  <!-- Force polyfill. -->
+  <!-- Uncomment this script to force the polyfill installation
   <script>
-    if (false) {
-      delete window.IntersectionObserver;
-      delete window.IntersectionObserverEntry;
-    }
+    delete window.IntersectionObserver;
+    delete window.IntersectionObserverEntry;
   </script>
+  -->
 
   <!-- Script under test -->
   <script src="intersection-observer.js"></script>
--- a/polyfill/intersection-observer-test.js
+++ b/polyfill/intersection-observer-test.js
@@ -953,18 +953,19 @@ describe('IntersectionObserver', function() {
       iframe.style.top = '0px';
       iframe.style.width = '100px';
       iframe.style.height = '200px';
-      iframe.srcdoc =
-        '<!DOCTYPE html><html><body>' +
-        '<style>' +
-        'body {margin: 0}' +
-        ' .target {height: 200px; margin-bottom: 2px; background: blue;}' +
-        '</style>';
       iframe.onerror = function() {
         done(new Error('iframe initialization failed'));
       };
       iframe.onload = function() {
         iframeWin = iframe.contentWindow;
         iframeDoc = iframeWin.document;
+        iframeDoc.open();
+        iframeDoc.write('<!DOCTYPE html><html><body>');
+        iframeDoc.write('<style>');
+        iframeDoc.write('body {margin: 0}');
+        iframeDoc.write('.target {height: 200px; margin-bottom: 2px; background: blue;}');
+        iframeDoc.write('</style>');
+        iframeDoc.close();
 
         function createTarget(id, bg) {
           var target = iframeDoc.createElement('div');
@@ -978,6 +979,7 @@ describe('IntersectionObserver', function() {
         iframeTargetEl2 = createTarget('target2', 'green');
         done();
       };
+      iframe.src = 'about:blank';
       rootEl.appendChild(iframe);
     });
 
--- a/polyfill/intersection-observer-test.js
+++ b/polyfill/intersection-observer-test.js
@@ -941,6 +941,652 @@ describe('IntersectionObserver', function() {
 
   });
 
+  describe('same-origin iframe', function() {
+    var iframe, win, doc;
+    var iframeTargetEl1, iframeTargetEl2;
+
+    beforeEach(function() {
+      iframe = document.createElement('iframe');
+      iframe.setAttribute('frameborder', '0');
+      iframe.setAttribute('scrolling', 'yes');
+      iframe.style.position = 'fixed';
+      iframe.style.top = '0px';
+      iframe.style.width = '100px';
+      iframe.style.height = '200px';
+      iframe.srcdoc =
+        '<!DOCTYPE html><html><body>' +
+        '<style>' +
+        'body {margin: 0}' +
+        ' .target {height: 200px; margin-bottom: 2px; background: blue;}' +
+        '</style>';
+      return new Promise(function(resolve, reject) {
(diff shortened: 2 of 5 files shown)


polyfill/intersection-observer.js
```diff
@@ -693,6 +793,28 @@ function getEmptyRect() {
   };
 }
 
+
+/**
+ * Inverts the intersection and bonding rect from the parent (frame) BCR to
```
philipwalton, line 798: s/bonding/bounding/

polyfill/intersection-observer-test.html
```diff
@@ -27,6 +27,14 @@
   <!-- Setup -->
   <script>mocha.setup({ui:'bdd'});</script>
 
+  <!-- Force polyfill. -->
+  <script>
+    if (false) {
```
philipwalton, line 32: Did you mean to keep this here?

polyfill/intersection-observer.js
dvoytenko: done

dvoytenko:
> Thanks for filing this PR. These changes look pretty reasonable on first look (with a few nits). Have you looked into added a test for this new behavior?

Definitely. I added several tests in the `describe('same-origin iframe')`. I added tests, for three categories of things:
 1. Monitoring scrolling/mutations/etc for iframes. Similar to non-iframed case.
 2. Calculation of intersectionRect and other entry structs.
 3. Main-vs-iframe-context coordination tests.

polyfill/intersection-observer-test.html
dvoytenko: Yes, but I can remove if you don't want it here. As it is, I thought of this as a documentation for how to quickly test the polyfill on a modern browser.

philipwalton:
Ahh, sorry, the GitHub UI was hiding the diff for that file, so I missed it. Will take a look later today.

polyfill/intersection-observer.js
```diff
@@ -262,58 +266,138 @@ IntersectionObserver.prototype._parseRootMargin = function(opt_rootMargin) {
...
 
-    clearInterval(this._monitoringInterval);
-    this._monitoringInterval = null;
+  var rootDoc = (this.root && this.root.ownerDocument || document);
 
-    removeEvent(window, 'resize', this._checkForIntersections, true);
-    removeEvent(document, 'scroll', this._checkForIntersections, true);
+  // Check if any dependent targets are still remaining.
+  var dependentTargets =
+      this._observationTargets.filter(function(item) {
```
jridgewell, line 350: We can use `_observationTargets.some()` to allow early exiting iteration.

polyfill/intersection-observer.js
```diff
@@ -262,58 +266,138 @@ IntersectionObserver.prototype._parseRootMargin = function(opt_rootMargin) {
...
 };
 
 
+/**
+ * Stops polling for intersection changes.
+ * @param {!Document} doc
+ * @private
+ */
+IntersectionObserver.prototype._unmonitorAllIntersections = function() {
+  var unsubscribes = [].concat(this._monitoringUnsubscribes);
```
jridgewell, line 392: Nit: Use `_monitoringUnsubscribes.slice()`

polyfill/intersection-observer.js
dvoytenko: Oops. Definitely!

polyfill/intersection-observer-test.html
philipwalton: I think it's fine to keep it here (to explain how to test, as you say) but I'd remove the `if (false)` and replace it with a comment that says something along the lines of: "comment out the below script to test the default behavior" (or something like that).

polyfill/intersection-observer.js
```diff
@@ -734,6 +861,21 @@ function getParentNode(node) {
 }
 
 
+/**
+ * Returns the embedding framem element, if any.
```
philipwalton, line 865: s/framem/frame/

polyfill/intersection-observer-test.js
```diff
@@ -941,6 +941,652 @@ describe('IntersectionObserver', function() {
 
   });
 
+  describe('same-origin iframe', function() {
```
philipwalton, line 944: Can you move these tests inside the `observe()` describe block. The structure of the tests currently uses a single describe block per method, so I think it makes sense to stick with that unless you think these tests really don't fit that pattern.

polyfill/intersection-observer-test.js
```diff
@@ -941,6 +941,652 @@ describe('IntersectionObserver', function() {
...
+      iframe.style.top = '0px';
+      iframe.style.width = '100px';
+      iframe.style.height = '200px';
+      iframe.srcdoc =
+        '<!DOCTYPE html><html><body>' +
+        '<style>' +
+        'body {margin: 0}' +
+        ' .target {height: 200px; margin-bottom: 2px; background: blue;}' +
+        '</style>';
+      return new Promise(function(resolve, reject) {
```
philipwalton, line 962: When this polyfill was written (years ago) it was designed to work all the way back to IE7.

Using `Promise` here would break the tests when run in IE, but I think you could get away with still using it by adding that polyfill to the polyfills the test runner loads: https://github.com/w3c/IntersectionObserver/blob/master/polyfill/intersection-observer-test.html#L22

polyfill/intersection-observer-test.js
```diff
@@ -941,6 +941,652 @@ describe('IntersectionObserver', function() {
...
+          to.eql(rect(iframeDoc.body.getBoundingClientRect()));
+
+        done();
+        io.disconnect();
+      }, {root: iframeDoc.body});
+      io.observe(iframeTargetEl1);
+      io.observe(iframeTargetEl2);
+    });
+
+    it('calculates rects for a fully visible frame', function(done) {
```
philipwalton, line 1062: Can you confirm that this test is passing for you? I see failures when running it both with and without the polyfill. (Note: I see the same for the tests whose description starts with "calculates..."):

Screen Shot 2020-04-08 at 5 44 12 PM

polyfill/intersection-observer-test.html
dvoytenko: Done.

polyfill/intersection-observer-test.js
dvoytenko: The tests are more orthogonal in this case and set up is very special for iframes and I wanted to reduce the spread across the board. But I also I wanted to make it concentrated to communicate "here are all the things that are special about iframes".

polyfill/intersection-observer-test.js
dvoytenko: Good point. I removed the promise. The rest of tests are using `done` callback, so this one should do as well.

polyfill/intersection-observer-test.js
dvoytenko: This is very odd. And especially w.r.t. to `right: 85`. But I can confirm that all tests pass for me for both native and polyfill and on all the browsers that I've tried. And I've tested on Chrome/Firefox/Safari. Having difficulties testing on IE at the moment, but will certainly find a way to test IE before merging (and any advise would be appreciated).

dvoytenko: Turns out, this is not odd at all. Just the test wasn't accounting for the scrollbar. I fixed it. And tested on IE11 and IE10.

dvoytenko:
@philipwalton I fixed the tests and tested with IE11 and IE10. PTAL.

polyfill/intersection-observer-test.js
philipwalton: OK. I still think it'd be nice if there were a way to logically group the same-document and iframe tests (maybe run all the tests in both scenarios), but since that would likely require a larger refactor, I don't want to block adding this feature.

dvoytenko:
Close/reopen to run CI again.

PR #421, opened 2020-04-15
Support 3p mode for the polyfill

dvoytenko:
This pull request enables the polyfill to support cross-origin intersection observer features, assuming the host page messages the correct intersection information.

### DIFF
--- a/polyfill/README.md
+++ b/polyfill/README.md
@@ -117,7 +117,7 @@ cross-origin iframe element. It can either use the this polyfill or any other
 approach. For each IntersectionObserverEntry for the iframe it will forward
 intersection data to the iframe via messaging.
 3. The iframe will load the polyfill and configure it by calling the
-`polyfillSetupCrossOriginUpdater()` method. It will call the provided callback
+`_setupCrossOriginUpdater()` method. It will call the provided callback
 whenever it receives the intersection data from the the parent via messaging.
 
 A hypothetical host code:
@@ -144,8 +144,8 @@ A hypothetical iframe code:
 
 ```javascript
 createMessagingChannel(parent, function(port) {
-  if (IntersectionObserver.polyfillSetupCrossOriginUpdater) {
-    var crossOriginUpdater = IntersectionObserver.polyfillSetupCrossOriginUpdater();
+  if (IntersectionObserver._setupCrossOriginUpdater) {
+    var crossOriginUpdater = IntersectionObserver._setupCrossOriginUpdater();
     port.onmessage = function(event) {
       crossOriginUpdater(
         deserialize(event.data.boundingClientRect),
--- a/polyfill/README.md
+++ b/polyfill/README.md
@@ -97,7 +97,7 @@ io.USE_MUTATION_OBSERVER = false;
 This is recommended in cases where the DOM will update frequently but you know those updates will have no affect on the position or your target elements.
 
 
-## IFRAME support
+## iframe support
 
 ### Same-origin iframes
 
@@ -113,12 +113,12 @@ The setup is as following:
 
 1. The host and iframe will establish a messaging channel.
 2. The host will setup its own IntersectionObserver instance for the
-cross-orgin iframe element. It can either use the this polyfill or any other
+cross-origin iframe element. It can either use the this polyfill or any other
 approach. For each IntersectionObserverEntry for the iframe it will forward
 intersection data to the iframe via messaging.
 3. The iframe will load the polyfill and configure it by calling the
 `polyfillSetupCrossOriginUpdater()` method. It will call the provided callback
-whenever it receives the intersection data from the the parent via messsaging.
+whenever it receives the intersection data from the the parent via messaging.
 
 A hypothetical host code:
 
--- a/polyfill/README.md
+++ b/polyfill/README.md
@@ -130,11 +130,15 @@ function forwardIntersectionToIframe(iframe) {
         boundingClientRect: serialize(boundingClientRect),
         intersectionRect: serialize(intersectionRect)
       });
-    });
+    }, {threshold: [0, 0.1, ..., 1]});
     io.observe(iframe);
   });
 }
 ```
+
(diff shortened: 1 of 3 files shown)


polyfill/intersection-observer.js
```diff
@@ -48,6 +48,20 @@ var document = window.document;
  */
 var registry = [];
 
+/**
+ * The signal updater for cross-origin intersection. When available, it means
```
philipwalton, line 52: Insteading of saying "when available" maybe say "when set" or "when not null"? To me "available" is a bit unclear.

polyfill/intersection-observer.js
```diff
@@ -48,6 +48,20 @@ var document = window.document;
...
+ * that the polyfill is configured to work in a cross-origin mode.
+ * @type {function(DOMRect, DOMRect)}
+ */
+var crossOriginUpdater = null;
+
+/**
+ * The current cross-origin intersection. Only available in the cross-origin
```
philipwalton, line 59: Same comment as above w.r.t. "available".

polyfill/intersection-observer.js
```diff
@@ -147,6 +161,45 @@ IntersectionObserver.prototype.POLL_INTERVAL = null;
 IntersectionObserver.prototype.USE_MUTATION_OBSERVER = true;
 
 
+/**
+ * Set up the polyfill in the cross-origin mode. The result is the
```
philipwalton, line 165: s/Set/Sets/

polyfill/intersection-observer.js
```diff
@@ -147,6 +161,45 @@ IntersectionObserver.prototype.POLL_INTERVAL = null;
...
+ * updater function that accepts two arguments: `boundingClientRect` and
+ * `intersectionRect` - just as these fields would be available to the
+ * parent via `IntersectionObserverEntry`. This function should be called
+ * each time the iframe receives intersection information from the parent
+ * window, e.g. via messaging.
+ * @return {function(DOMRect, DOMRect)}
+ */
+IntersectionObserver.polyfillSetupCrossOriginUpdater = function() {
```
philipwalton, line 173: I'm not crazy about the use of the `polyfill` prefix here. I know you're trying to avoid collisions with the native API, but this kinda makes it sound like the method is polyfilling something.

How would you feel about `_setupCrossOriginUpdater()` or `__setupCrossOriginUpdater()`?

(BTW, I know I'm already kinda breaking this with `POLL_INTERVAL` etc., but those feel a bit safer since they're all caps. Arguably they should also be `__POLL_INTERVAL` though...)

polyfill/intersection-observer-test.js
```diff
@@ -941,8 +941,9 @@ describe('IntersectionObserver', function() {
 
   });
 
-  describe('same-origin iframe', function() {
+  describe('iframe', function() {
     var iframe, win, doc;
```
philipwalton, line 945: Didn't notice this before, but my editor is telling me `win` and `doc` are not used here.

polyfill/intersection-observer-test.js
```diff
@@ -969,6 +970,12 @@ describe('IntersectionObserver', function() {
         iframeDoc.write('</style>');
         iframeDoc.close();
 
+        // Ensure the documentElement and body are always sorted on top.
```
philipwalton, line 973: Not sure I follow this comment. Can you explain what this is ensuring?

polyfill/README.md
```diff
@@ -96,6 +96,67 @@ io.USE_MUTATION_OBSERVER = false;
 
 This is recommended in cases where the DOM will update frequently but you know those updates will have no affect on the position or your target elements.
 
+
+## IFRAME support
```
philipwalton, line 100: I'd make IFRAME all lower case here.

polyfill/README.md
```diff
@@ -96,6 +96,67 @@ io.USE_MUTATION_OBSERVER = false;
...
+### Cross-origin iframes
+
+Additional code and configuration are required to support cross-origin iframes,
+both on the iframe and host sides.
+
+The setup is as following:
+
+1. The host and iframe will establish a messaging channel.
+2. The host will setup its own IntersectionObserver instance for the
+cross-orgin iframe element. It can either use the this polyfill or any other
```
philipwalton, line 116: s/orgin/origin/

polyfill/README.md
```diff
@@ -96,6 +96,67 @@ io.USE_MUTATION_OBSERVER = false;
...
+approach. For each IntersectionObserverEntry for the iframe it will forward
+intersection data to the iframe via messaging.
+3. The iframe will load the polyfill and configure it by calling the
+`polyfillSetupCrossOriginUpdater()` method. It will call the provided callback
+whenever it receives the intersection data from the the parent via messsaging.
```
philipwalton, line 121: s/messsaging/messaging/

polyfill/intersection-observer-test.js
dvoytenko: I was just trying to follow `sortRecords` function that was available in these tests. It's used to make records order predictable, which totally makes sense. So I wanted to make sure that if `<html>` or `<body>` are observed, they'd always be sorted in the predictable order as well, along with all the other targets.

polyfill/intersection-observer.js
dvoytenko: I'd like that, but I'm a little freaked out by the possibility that some compilers/bundles will consider it a module-boundary-private and obfuscate. However, the intent of this function is to be called outside of this module. I run into such things occasionally and so rather careful about cross-module function names with a head/tail underscore.

polyfill/intersection-observer-test.js
dvoytenko: I added the comment in the code about `sortRecords`.

polyfill/intersection-observer.js
dvoytenko: Renamed to `_setupCrossOriginUpdater`.

PR #423, opened 2020-04-23
Polyfill: ensure the complete DOMRect shape is returned

dvoytenko:
Context:

* Most of browsers use `DOMRect` structure with the shape `{x, y, top, left, width, height, bottom, right}`. The `x` and `y` simply mirror `top` and `left` when `width` and `height` are positive, which should be the case for us always.
* IE has its own `ClientRect` with the same `{top, left, width, height, bottom, right}`.
* I believe the polyfill always uses the native result of the `getBoundingClientRect()` for `Entry.boundingClientRect`. However it calculates `rootBounds` and `intersectionRect` fields.
* Consequently, I think the biggest possible source of compatibility issues is for the fields we calculate that currently always miss `x` and `y`. For instance [this computation](https://github.com/w3c/IntersectionObserver/blob/9bb05eaf228d98d807f20006cfa8de1dcc56850b/polyfill/intersection-observer.js#L806). Or event a simple [getEmptyRect](https://github.com/w3c/IntersectionObserver/blob/9bb05eaf228d98d807f20006cfa8de1dcc56850b/polyfill/intersection-observer.js#L854).
* Thus, on the platforms that support `DOMRect`, we definitely have an inconsistency, since the `boundingClientRect` has `x/y` and the other two rects do not.

I could have updated only the polyfill calculations only. But for consistency, I decided to just update "fill in" x/y in all rects returned by the `IntersectionObserverEntry`.

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -71,9 +71,9 @@ var crossOriginRect = null;
 function IntersectionObserverEntry(entry) {
   this.time = entry.time;
   this.target = entry.target;
-  this.rootBounds = entry.rootBounds;
-  this.boundingClientRect = entry.boundingClientRect;
-  this.intersectionRect = entry.intersectionRect || getEmptyRect();
+  this.rootBounds = fixDOMRect(entry.rootBounds);
+  this.boundingClientRect = fixDOMRect(entry.boundingClientRect);
+  this.intersectionRect = fixDOMRect(entry.intersectionRect || getEmptyRect());
   this.isIntersecting = !!entry.intersectionRect;
 
   // Calculates the intersection ratio.
@@ -859,6 +859,30 @@ function getEmptyRect() {
     right: 0,
     width: 0,
     height: 0
+  };
+}
+
+
+/**
+ * Ensure that the result has all of the necessary fields of the DOMRect.
+ * Specifically this ensures that `x` and `y` fields are set.
+ *
+ * @param {?DOMRect} rect
+ * @return {?DOMRect}
+ */
+function fixDOMRect(rect) {
+  if (!rect || 'x' in rect) {
+    return rect;
+  }
+  return {
+    top: rect.top,
+    y: rect.top,
+    bottom: rect.bottom,
+    left: rect.left,
+    x: rect.left,
+    right: rect.right,
+    width: rect.width,
+    height: rect.height
   };
 }
 
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -51,13 +51,13 @@ var registry = [];
 /**
  * The signal updater for cross-origin intersection. When not null, it means
  * that the polyfill is configured to work in a cross-origin mode.
- * @type {function(DOMRect, DOMRect)}
+ * @type {function(DOMRect|ClientRect, DOMRect|ClientRect)}
  */
 var crossOriginUpdater = null;
 
 /**
  * The current cross-origin intersection. Only used in the cross-origin mode.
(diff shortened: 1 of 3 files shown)


polyfill/intersection-observer.js
```diff
@@ -863,6 +863,30 @@ function getEmptyRect() {
...
 
+/**
+ * Ensure that the result has all of the necessary fields of the DOMRect.
+ * Specifically this ensures that `x` and `y` fields are set.
+ *
+ * @param {?DOMRect} rect
+ * @return {?DOMRect}
+ */
+function fixDOMRect(rect) {
+  if (!rect || 'x' in rect) {
```
philipwalton, line 874: Can you clarify what this check is for? Maybe add a comment?

polyfill/intersection-observer.js
```diff
@@ -863,6 +863,30 @@ function getEmptyRect() {
...
+    return rect;
+  }
+  return {
+    top: rect.top,
+    y: rect.top,
```
philipwalton, line 879: Do some browsers not support the `y` property? Why is this mirroring the `top` value?

polyfill/intersection-observer.js
```diff
@@ -863,6 +863,30 @@ function getEmptyRect() {
...
 
 
+/**
+ * Ensure that the result has all of the necessary fields of the DOMRect.
+ * Specifically this ensures that `x` and `y` fields are set.
+ *
+ * @param {?DOMRect} rect
+ * @return {?DOMRect}
+ */
+function fixDOMRect(rect) {
```
philipwalton, line 873: Maybe `validateDOMRect()`, since it's not necessarily fixing anything, right?

polyfill/intersection-observer.js
dvoytenko: Let me describe context here and we'll discuss what's the best path forward.

* Most of browsers use `DOMRect` structure with the shape `{x, y, top, left, width, height, bottom, right}`. The `x` and `y` simply mirror `top` and `left` when `width` and `height` are positive, which should be the case for us always.
* IE has its own `ClientRect` with the same `{top, left, width, height, bottom, right}`.
* I believe the polyfill always uses the native result of the `getBoundingClientRect()` for `Entry.boundingClientRect`. However it calculates `rootBounds` and `intersectionRect` fields.
* Consequently, I think the biggest possible source of compatibility issues is for the fields we calculate that currently always miss `x` and `y`. For instance [this computation](https://github.com/w3c/IntersectionObserver/blob/9bb05eaf228d98d807f20006cfa8de1dcc56850b/polyfill/intersection-observer.js#L806). Or event a simple [getEmptyRect](https://github.com/w3c/IntersectionObserver/blob/9bb05eaf228d98d807f20006cfa8de1dcc56850b/polyfill/intersection-observer.js#L854).
* Thus, on the platforms that support `DOMRect`, we definitely have an inconsistency, since the `boundingClientRect` has `x/y` and the other two rects do not.

I could have updated only the polyfill calculations only. But for consistency, I decided to just update "fill in" x/y in all rects returned by the `IntersectionObserverEntry`.

polyfill/intersection-observer.js
dvoytenko: Done

polyfill/intersection-observer.js
dvoytenko: Updated comments.

polyfill/intersection-observer.js
dvoytenko: Changed to ensureDOMRect.

PR #425, opened 2020-04-23
[polyfill] get correct parent node for slot in shadow dom

### LINKED ISSUE #424: Polyfill: getParentNode returns incorrect result if target was added as a child through a shadow dom slot
Take a look at description here https://github.com/w3c/IntersectionObserver/issues/303#issuecomment-395948794

ennjin:
Closes #424

### DIFF
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -965,16 +965,17 @@ function getParentNode(node) {
     return getFrameElement(node);
   }
 
+  // If the parent has element that is assigned through shadow root slot
+  if (parent && parent.assignedSlot) {
+    parent = parent.assignedSlot.parentNode
+  }
+
   if (parent && parent.nodeType == 11 && parent.host) {
     // If the parent is a shadow root, return the host element.
     return parent.host;
   }
 
   if (parent && parent.assignedSlot) {
-    // If the parent is assigned into shadow root slot
-    if (parent.assignedSlot.parentNode.host) {
-      return parent.assignedSlot.parentNode.host;
-    }
     // If the parent is distributed in a <slot>, return the parent of a slot.
     return parent.assignedSlot.parentNode;
   }
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -971,6 +971,10 @@ function getParentNode(node) {
   }
 
   if (parent && parent.assignedSlot) {
+    // If the parent is assigned into shadow root slot
+    if (parent.assignedSlot.parentNode.host) {
+      return parent.assignedSlot.parentNode.host;
+    }
     // If the parent is distributed in a <slot>, return the parent of a slot.
     return parent.assignedSlot.parentNode;
   }
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -975,11 +975,6 @@ function getParentNode(node) {
     return parent.host;
   }
 
-  if (parent && parent.assignedSlot) {
-    // If the parent is distributed in a <slot>, return the parent of a slot.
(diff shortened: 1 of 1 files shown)


ennjin:
@szager-chromium, @philipwalton, @dvoytenko someone can take a look at this?

polyfill/intersection-observer.js
```diff
@@ -971,6 +971,10 @@ function getParentNode(node) {
   }
 
   if (parent && parent.assignedSlot) {
+    // If the parent is assigned into shadow root slot
```
dvoytenko, line 974: This sentence seems incomplete.

polyfill/intersection-observer.js
```diff
@@ -971,6 +971,10 @@ function getParentNode(node) {
...
+    if (parent.assignedSlot.parentNode.host) {
```
dvoytenko, line 975: It might be better to move this part before the previous `if (parent.nodeType == 11 && parent.host)` and do something like:

```
if (parent && parent.assignedSlot) {
  parent = parent.assignedSlot.parentNode;
}
```

Then the `if (parent.nodeType == 11 && parent.host)` will do the right thing.

polyfill/intersection-observer.js
```diff
@@ -965,6 +965,11 @@ function getParentNode(node) {
     return getFrameElement(node);
   }
 
+  // If the parent has element that is assigned through shadow root slot
+  if (parent && parent.assignedSlot) {
```
dvoytenko, line 969: This looks good, but it also means that the second `if (parent && parent.assignedSlot) {` at line 978 is no longer needed, right?

ennjin: Yeah, makes sense.

ennjin:
Hm, is it ready to merge?

dvoytenko (approved):

dvoytenko:
This change looks good to me. The owners of the repo still need to review and merge.

ennjin:
@emilio it would be nice if you can check this :+1:

polyfill/intersection-observer.js
```diff
@@ -965,16 +965,16 @@ function getParentNode(node) {
...
+    parent = parent.assignedSlot.parentNode
```
emilio, line 970: I don't think this is correct if the assigned slot has a box (via e.g. `slot { display: block }`).

polyfill/intersection-observer.js
```diff
@@ -965,16 +965,16 @@ function getParentNode(node) {
     return getFrameElement(node);
   }
 
+  // If the parent has element that is assigned through shadow root slot
+  if (parent && parent.assignedSlot) {
```
emilio, line 969: Also probably this needs to be a `while (parent && parent.assignedSlot)` to handle nested slots.

emilio (commented):
I'm curious what is the point of fixing up the polyfill if modern browsers all support the API correctly.

ennjin:
> I'm curious what is the point of fixing up the polyfill if modern browsers all support the API correctly.

Actually there are customers using legacy browsers like chrome <= 50 and safari <= 11 :smile_cat:

polyfill/intersection-observer.js
ennjin: > I don't think this is correct if the assigned slot has a box (via e.g. slot { display: block }).

Not sure if anyone style slot element

emilio (approved):
I mean, this is more correct than the existing code I guess. But we should really move the polyfill to a separate repo IMO.

ennjin:
@emilio Thanks for review. But I have a question, when the following changes will be bumped to npm?

Issue #430, opened 2020-05-11
When an Element is observed, ensure that it's document runs the "update the rendering" steps at the next rendering opportunity.


szager-chromium:
The expectation of developers is that when observer.observe(target) is called, an initial notification will be generated during the next rendering update (step 11 in the [Document Processing Model](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-processing-model)). However, step 11.4 in the processing model allows an implementation to skip processing for a document under certain circumstances:

> Unnecessary rendering: Remove from docs all Document objects which meet both of the following conditions:
>
> - The user agent believes that updating the rendering of the Document's browsing context would have no visible effect, and
> - The Document's map of animation frame callbacks is empty.

So it's possible that an initial IntersectionObserver notification will **not** be generated at the next rendering opportunity. This issue was first discussed in previous [issue](https://github.com/w3c/IntersectionObserver/issues/165), and it was agreed that creating an initial notification at the first opportunity was a good idea. The spec change associated with that bug did not, however, anticipate or account for the "unnecessary rendering" mechanism described above.

The spec should be amended to include an additional, mandatory condition for skipping processing for a document; something like:

> There is no IntersectionObserver whose root is in the DOM tree of the Document, which is observing a target Element for which no notification has been generated since the time observe() was called.

The actual language should probably refer to the IntersectionObserver's [ObservationTargets](https://w3c.github.io/IntersectionObserver/#dom-intersectionobserver-observationtargets-slot) slot, the target Element's [Registered IntersectionObservers](https://w3c.github.io/IntersectionObserver/#dom-element-registeredintersectionobservers-slot) slot, and the *previousThresholdIndex* property on **IntersectionObserverRegistration**. It will probably be kinda gross, so the full details should probably be in the IntersectionObserver spec, linked from the "unnecessary rendering" section of the processing model spec.

PR #445, opened 2020-09-04
Force rendering update for pending initial targets

szager-chromium:
Issue #430

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/pull/445.html" title="Last updated on Sep 5, 2020, 4:44 PM UTC (f2e6efb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/445/64583b6...f2e6efb.html" title="Last updated on Sep 5, 2020, 4:44 PM UTC (f2e6efb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -13,7 +13,7 @@ Abstract: This specification describes  an API that can be used to understand th
 Group: webapps
 Repository: W3C/IntersectionObserver
 Test Suite: http://w3c-test.org/intersection-observer/
-Ignored Vars: rootMargin, docs
+Ignored Vars: rootMargin, docs, now
 </pre>
 
 <pre class="anchors">
@@ -38,7 +38,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
 		type: dfn; text: queue a task
 		type: dfn; text: run the fullscreen rendering steps
 		type: dfn; text: run the animation frame callbacks
-		url: #processing-model-8; type: dfn; text: HTML Processing Model
+		url: #event-loop-processing-model; type: dfn; text: HTML Processing Model
 	urlPrefix: infrastructure.html;
 		type: dfn; text: rules for parsing dimension values
 		url: #dfn-callback-this-value; type: dfn; text: callback this value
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -669,15 +670,31 @@ HTML Processing Model: Event Loop</h4>
...
 
+<h4 id='pending-initial-observation'>Pending initial IntersectionObserver targets</h4>
+A {{document}} is said to have <dfn>pending initial IntersectionObserver targets</dfn>
+if there is at least one {{IntersectionObserver}} meeting these criteria:
+	<ol>
+		<li>The |observer|'s {{IntersectionObserver|root}} is in the |document|
+		   (for the <a>top-level browsing context</a>, this includes <a>implicit root</a> observers).
+		</li>
+		<li>The |observer| has at least one |target| in its {{[[ObservationTargets]]}} slot for which
+		    no {{IntersectionObserverEntry}} has yet been queued.
```
emilio, line 690: I'm not quite sure this is the right condition. This means that calling `observe()` on something that's off-screen would cause forced rendering updates all frames, right?

szager-chromium: Yesl; unless some of the frames could skip rendering without affecting the intersection calculation.

emilio: Right, I think that's wrong. The "Unnecessary rendering" section already has a condition on:

> The user agent believes that updating the rendering of the Document's browsing context would have no visible effect

Which adding this other condition would break, right? That is. IntersectionObserver needs to force-update the rendering _once_, to guarantee that if the element is visible the observer runs. If the element is not visible, there's no need to do that, and you can skip rendering updates as usual (until something causes you to do that).

szager-chromium: I'm not sure I understand you. Here's the problematic scenario (running on a page with unchanging layout):

```
setTimeout(() => {
  let target = document.getElementById('target');
  let observer = new IntersectionObserver(entries => { alert('Hello, world!'); });
  observer.observe(target);
}, 1000);
```

In this case, it is true that "the rendering of the Document's browsing context would have no visible effect." So the implementation may skip rendering updates indefinitely, and no notification will be generated because the IntersectionObserver steps of the document lifecycle don't run if rendering is skipped. The point of the issue is that we should generate a notification for the target at the next opportunity, whether the target is visible or not.

emilio: Ouch, yeah, sorry, I was misreading. This looks good, sorry.

emilio (approved):

PR #453, opened 2020-09-23
Allow 0 (without unit) as a valid offset for rootMargin

### LINKED ISSUE #244: rootMargin should allow 0 without units
having to write `0px` or `0%` seems unnecessary and is error-prone for developers used to type this way with CSS margin shorthand.

related #126

lebnic:
Closes #244 

Implementation commitment:

### DIFF
--- a/polyfill/intersection-observer-test.js
+++ b/polyfill/intersection-observer-test.js
@@ -91,7 +91,7 @@ describe('IntersectionObserver', function() {
       io = new IntersectionObserver(noop, {rootMargin: '10px 20% 0px'});
       expect(io.rootMargin).to.be('10px 20% 0px 20%');
 
-      io = new IntersectionObserver(noop, {rootMargin: '0px 0px -5% 5px'});
+      io = new IntersectionObserver(noop, {rootMargin: '0px 0 -5% 5px'});
       expect(io.rootMargin).to.be('0px 0px -5% 5px');
 
       // TODO(philipwalton): the polyfill supports fractional pixel and
--- a/polyfill/intersection-observer.js
+++ b/polyfill/intersection-observer.js
@@ -319,6 +319,9 @@ IntersectionObserver.prototype._initThresholds = function(opt_threshold) {
 IntersectionObserver.prototype._parseRootMargin = function(opt_rootMargin) {
   var marginString = opt_rootMargin || '0px';
   var margins = marginString.split(/\s+/).map(function(margin) {
+    if (margin === '0') {
+      return {value: 0, unit: 'px'};
+    }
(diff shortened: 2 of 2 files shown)


lebnic:
FYI @philipwalton

dontcallmedom:

emilio (commented):
This only fixes the polyfill, not the specification, right? What's the point of that?

lebnic:
> This only fixes the polyfill, not the specification, right? What's the point of that?

Documentation states that `[rootMargin] Can have values similar to the CSS margin property` (ref: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)

Thus, one would expect, for exemple, a value of `rootMargin: "200px 0 0"` to be valid. However, **current implementation will throw an error**, because only `0px` or `0%` are valid offsets, but not `0`.

Hence this PR.

Moreover, there has been an open issue (https://github.com/w3c/IntersectionObserver/issues/244) since 2017 for this particular use-case, so I thought that maybe my proposed changes would be beneficial to others. If not, I guess we could simply write `rootMargin: "200px 0px 0px"` and move on with our lives. Anyways, thanks for the review @emilio , and let me know if there is anything else I can clarify

emilio:
Well, my point is that:

 * Documentation, right now, is wrong. https://w3c.github.io/IntersectionObserver/#parse-a-root-margin disallows `0`. It also disallows `calc()` / `min()` / `max()` / viewport units and other relative units that you can use in the `margin` shorthand (some of them for very good reason, it's not clear what `em` would do here for example!).

 * You're changing a polyfill that doesn't have anything to do with any browser implementation. If you want browsers to change, first you need to change the spec (which is in this repo, and is in the `index.bs` file), then you need to change browsers (filing bugs in the relevant bug trackers).

So changing the polyfill without changing the specification and at the very least filing browser bugs and adding tests to wpt seems a bit odd / not useful.

lebnic:
> Well, my point is that:
> 
> * Documentation, right now, is wrong. https://w3c.github.io/IntersectionObserver/#parse-a-root-margin disallows `0`. It also disallows `calc()` / `min()` / `max()` / viewport units and other relative units that you can use in the `margin` shorthand (some of them for very good reason, it's not clear what `em` would do here for example!).
> * You're changing a polyfill that doesn't have anything to do with any browser implementation. If you want browsers to change, first you need to change the spec (which is in this repo, and is in the `index.bs` file), then you need to change browsers (filing bugs in the relevant bug trackers).
> 
> So changing the polyfill without changing the specification and at the very least filing browser bugs and adding tests to wpt seems a bit odd / not useful.

You are 100% right, my bad.

PR #455, opened 2020-09-23
Clarify invocation parameters for callback

### LINKED ISSUE #407: Missing argument or wrong link in algorithm for notifying observers
Step 3.4 in [3.2.2. Notify Intersection Observers](https://w3c.github.io/IntersectionObserver/#notify-intersection-observers-algo) says to call a callback with three arguments: the queue, the observer and an unspecified value with a type of ‘callback this value’. But the callback only accepts two arguments. Maybe the reference to a ‘callback this value’ was meant to be a reference to an algorithm instead, in particular [‘call a user object’s operation’](https://heycam.github.io/webidl/#call-a-user-objects-operation)?

szager-chromium:
Also remove unused external anchor "rules for parsing dimesion values".

Closes #407

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/pull/455.html" title="Last updated on Sep 30, 2020, 7:51 PM UTC (e7f24e3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/455/4c53ab2...e7f24e3.html" title="Last updated on Sep 30, 2020, 7:51 PM UTC (e7f24e3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -41,6 +41,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
 		type: dfn; text: relevant settings object
 		url: #event-loop-processing-model; type: dfn; text: HTML Processing Model
 		url: #concept-environment-top-level-origin; type: dfn; text: top-level origin
+	urlPrefix: infrastructure.html;
+		url: #dfn-callback-this-value; type: dfn; text: callback this value
 	urlPrefix: origin.html
 		type: dfn; text: origin
 urlPrefix: https://heycam.github.io/webidl/
@@ -565,7 +567,8 @@ run these steps:
 	2. Let |queue| be a copy of |observer|'s internal {{[[QueuedEntries]]}}
 		slot.
 	3. Clear |observer|'s internal {{[[QueuedEntries]]}} slot.
-	4. Invoke |callback| with |queue| as the first argument and |observer| as the second argument.
+	4. Invoke |callback| with |queue| as the first argument, |observer| as the second argument,
+		and |observer| as the <a>callback this value</a>.
 		If this throws an exception, <a>report the exception</a>.
 
 <h4 id='queue-intersection-observer-entry-algo'>
--- a/index.bs
+++ b/index.bs
@@ -41,9 +41,6 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
 		type: dfn; text: relevant settings object
 		url: #event-loop-processing-model; type: dfn; text: HTML Processing Model
 		url: #concept-environment-top-level-origin; type: dfn; text: top-level origin
-	urlPrefix: infrastructure.html;
-		type: dfn; text: rules for parsing dimension values
-		url: #dfn-callback-this-value; type: dfn; text: callback this value
 	urlPrefix: origin.html
 		type: dfn; text: origin
 urlPrefix: https://heycam.github.io/webidl/
@@ -568,8 +565,7 @@ run these steps:
 	2. Let |queue| be a copy of |observer|'s internal {{[[QueuedEntries]]}}
 		slot.
 	3. Clear |observer|'s internal {{[[QueuedEntries]]}} slot.
-	4. Invoke |callback| with |queue| as the first argument and |observer|
-		as the second argument and <a>callback this value</a>.
+	4. Invoke |callback| with |queue| as the first argument and |observer| as the second argument.
 		If this throws an exception, <a>report the exception</a>.
 
 <h4 id='queue-intersection-observer-entry-algo'>


index.bs
```diff
@@ -568,8 +565,7 @@ run these steps:
 	2. Let |queue| be a copy of |observer|'s internal {{[[QueuedEntries]]}}
 		slot.
 	3. Clear |observer|'s internal {{[[QueuedEntries]]}} slot.
-	4. Invoke |callback| with |queue| as the first argument and |observer|
-		as the second argument and <a>callback this value</a>.
+	4. Invoke |callback| with |queue| as the first argument and |observer| as the second argument.
```
emilio, line 568: I don't think this is right? The text was meant to say that the observer should be the callback `this` value as https://html.spec.whatwg.org/#custom-element-reactions:dfn-callback-this-value. This is making stuff less defined. I think it could probably be reworded like:

```
Invoke |callback| with |queue| and |observer| as arguments, and with |observer| as the <a>callback this value</a>.
```

Or something like that?

index.bs
```diff
@@ -41,9 +41,6 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
 		type: dfn; text: relevant settings object
 		url: #event-loop-processing-model; type: dfn; text: HTML Processing Model
 		url: #concept-environment-top-level-origin; type: dfn; text: top-level origin
-	urlPrefix: infrastructure.html;
-		type: dfn; text: rules for parsing dimension values
```
emilio, line 45: Removing the reference to `rules for parsing dimension values` is probably not intended?

emilio (changes requested):

index.bs
szager-chromium: I think this bit of the spec pre-dates my involvement, and I never looked at it closely before. Binding the observer as the 'this' argument doesn't make much sense, since the observer object is already passed as an argument. Chromium's implementation doesn't do this; does Firefox?

index.bs
szager-chromium: That reference is no longer used in the spec, this is just cleanup that snuck into this patch. I can remove it if you'd prefer a clean patch.

emilio: Either way is fine. Maybe note it in the commit message?

index.bs
emilio: What is the "this" value in Chromium? This seems to log the observer in both Firefox and Chromium:

```
new IntersectionObserver(function() { console.log("this", this) }).observe(document.documentElement)
```

szager-chromium: Oh, this is interesting -- the following behavior holds for both chromium and firefox:

> new IntersectionObserver(function() { console.log("this", this) }).observe(document.documentElement)
this
IntersectionObserver {root: null, rootMargin: "0px 0px 0px 0px", thresholds: Array(1), delay: 0, trackVisibility: false}

> new IntersectionObserver(() => { console.log("this", this) }).observe(document.documentElement)
this
Window {parent: Window, opener: null, top: Window, length: 4, frames: Window, …}

The difference, in case you can't spot it, is "function()" as opposed to "() =>". I'm not an ECMAScript expert by any stretch of the imagination, so I have no idea why it works that way. To add to the intrigue:

> new IntersectionObserver((function() { console.log("this.a", this.a) }).bind({a : "hello"})).observe(document.documentElement)
this.a
hello

> new IntersectionObserver((() => { console.log("this.a", this.a) }).bind({a : "hello"})).observe(document.documentElement)
this.a
undefined

I don't understand this at all, but it is at least reassuring that this mysterious behavior appears to be interoperable :) In the chromium code, we don't alter the 'this' binding of the callback at all, and I assume that must be true for firefox as well. I think that's the desired behavior; there's no reason to alter the 'this' binding.

index.bs
szager-chromium: Will do.

index.bs
szager-chromium: OK, I got our resident ECMAScript expert to explain this to me. It appears that both chromium and firefox *are* binding the IntersectionObserver instance to the callback 'this' argument. In the interest of backwards compatibility, we probably shouldn't change this.

I don't then see any problem with the existing spec language, so I'm inclined to just abandon this change, unless you think the existing language could be improved somehow.

emilio: Yeah, arrow functions capture the `this` value from the outer scope, that's the difference and why you need to use `function() {}` to see this.

Yeah, I'm ok with either closing this, or maybe adding a few commas to clarify that the `this` value is something different, no big deal either way.

szager-chromium: Done.

index.bs
szager-chromium: Done.

emilio (approved):
Thanks!

PR #459, opened 2020-10-13
Specify task source for IntersectionObserver notification tasks

### LINKED ISSUE #402: Task source is not specified
In [Queue an Intersection Observer task](https://w3c.github.io/IntersectionObserver/#queue-intersection-observer-task), we queue a task to the document's event loop, however to queue a task you must also specify a task source, which we do not.

Chromium seems to use a [special task source](https://cs.chromium.org/chromium/src/third_party/blink/public/platform/task_type.h?l=204&rcl=fccde1d4f480610da9fe5e95a91efcbb721d2730) specific to Intersection Observer for this, which I think is probably the most extensible solution. So the spec should probably define this task source and use it appropriately.

szager-chromium:
Closes #402

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/pull/459.html" title="Last updated on Oct 13, 2020, 6:54 PM UTC (024e507)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/459/89aacf7...024e507.html" title="Last updated on Oct 13, 2020, 6:54 PM UTC (024e507)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -564,13 +564,18 @@ and an {{Element}} |target|, follow these steps:
 <h4 id='queue-intersection-observer-task'>
 Queue an Intersection Observer Task</h4>
 
+The <dfn>IntersectionObserver task source</dfn> is a [=task source=] used for
+scheduling tasks to [[#notify-intersection-observers-algo]].
+
 To <dfn>queue an intersection observer task</dfn> for a {{document}} |document|,
 run these steps:
 
 1. If |document|'s <a>IntersectionObserverTaskQueued</a> flag is set to true,
 	return.
 2. Set |document|'s <a>IntersectionObserverTaskQueued</a> flag to true.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -589,7 +594,8 @@ run these steps:
 	2. Let |queue| be a copy of |observer|'s internal {{[[QueuedEntries]]}}
 		slot.
 	3. Clear |observer|'s internal {{[[QueuedEntries]]}} slot.
-	4. Invoke |callback| with |queue| as the first argument, |observer| as the second argument,
+	4. Let |callback| be the value of |observer|'s internal {{[[callback]]}} slot.
```
emilio, line 597: This looks fine though it may be worth a separate commit.

emilio (approved):

PR #465, opened 2020-11-13
Support iframe viewport tracking

samouri:
**summary**
Implements iframe viewport tracking for the polyfill, as described in https://github.com/w3c/IntersectionObserver/issues/372#issuecomment-575287674.

Implementation already completed:

 * [x] WebKit (https://bugs.webkit.org/show_bug.cgi?id=208047)
 * [x] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=1015183)
 * [x] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=1617154)

cc @dvoytenko

### DIFF
--- a/polyfill/intersection-observer-test.js
+++ b/polyfill/intersection-observer-test.js
@@ -69,12 +69,15 @@ describe('IntersectionObserver', function() {
       io = new IntersectionObserver(noop);
       expect(io.root).to.be(null);
 
+      io = new IntersectionObserver(noop, {root: document});
+      expect(io.root).to.be(document);
+
       io = new IntersectionObserver(noop, {root: rootEl});
-      expect(io.root).to.be(rootEl);
+      expect(io.root).to.be(rootEl); 
     });
 
 
-    it('throws when root is not an Element', function() {
+    it('throws when root is not a Document or Element', function() {
       expect(function() {
         io = new IntersectionObserver(noop, {root: 'foo'});
       }).to.throwException();
--- a/polyfill/intersection-observer-test.js
+++ b/polyfill/intersection-observer-test.js
@@ -1401,6 +1401,31 @@ describe('IntersectionObserver', function() {
         io.observe(iframeTargetEl2);
       });
 
+      it('handles tracking iframe viewport', function(done) {
+        iframe.style.height = '100px';
+        // {root:iframeDoc} means to track the iframe viewport irrespective of toplevel viewport
+        var io = new IntersectionObserver(
+          function (records) {
+            var subviewportWidth =
+              iframeDoc.documentElement.clientWidth || iframeDoc.body.clientWidth;
+            var subviewportHeight =
+              iframeDoc.documentElement.clientHeight || iframeDoc.body.clientHeight;
+
+            expect(records.length).to.be(1);
+            expect(records[0].rootBounds.top).to.be(0);
+            expect(records[0].rootBounds.left).to.be(0);
+            expect(records[0].rootBounds.right).to.be(subviewportWidth);
+            expect(records[0].rootBounds.width).to.be(subviewportWidth);
+            expect(records[0].rootBounds.bottom).to.be(subviewportHeight);
+            expect(records[0].rootBounds.height).to.be(subviewportHeight);
+            done();
+          },
+          { root: iframeDoc }
+        );
+
+        io.observe(iframeTargetEl1);
+      });
+
       it('handles style changes', function(done) {
         var spy = sinon.spy();
 
@@ -1627,10 +1652,10 @@ describe('IntersectionObserver', function() {
       var ASYNC_TIMEOUT = 300;
 
       beforeEach(function(done) {
-        /* Uncomment these lines to force polyfill inside the iframe.
+        // /* Uncomment these lines to force polyfill inside the iframe.
         delete iframeWin.IntersectionObserver;
         delete iframeWin.IntersectionObserverEntry;
-        */
+        // */
 
         // Install polyfill right into the iframe.
         if (!iframeWin.IntersectionObserver) {
@@ -2290,10 +2315,10 @@ describe('IntersectionObserver', function() {
       beforeEach(function(done) {
         Object.defineProperty(iframeWin, 'frameElement', {value: null});
 
-        /* Uncomment these lines to force polyfill inside the iframe.
+        // /* Uncomment these lines to force polyfill inside the iframe.
         delete iframeWin.IntersectionObserver;
         delete iframeWin.IntersectionObserverEntry;
-        */
+        // */
 
         // Install polyfill right into the iframe.
         if (!iframeWin.IntersectionObserver) {
@@ -3024,6 +3049,31 @@ describe('IntersectionObserver', function() {
             done();
           }
         ], done);
+      });
+
+      it('handles tracking iframe viewport', function(done) {
+        iframe.style.height = '100px';
+        // {root:iframeDoc} means to track the iframe viewport irrespective of toplevel viewport
+        var io = new IntersectionObserver(
+          function (records) {
+            var subviewportWidth =
+              iframeDoc.documentElement.clientWidth || iframeDoc.body.clientWidth;
+            var subviewportHeight =
+              iframeDoc.documentElement.clientHeight || iframeDoc.body.clientHeight;
+
+            expect(records.length).to.be(1);
+            expect(records[0].rootBounds.top).to.be(0);
+            expect(records[0].rootBounds.left).to.be(0);
+            expect(records[0].rootBounds.right).to.be(subviewportWidth);
+            expect(records[0].rootBounds.width).to.be(subviewportWidth);
+            expect(records[0].rootBounds.bottom).to.be(subviewportHeight);
+            expect(records[0].rootBounds.height).to.be(subviewportHeight);
+            done();
+          },
+          { root: iframeDoc }
+        );
+
+        io.observe(iframeTargetEl1);
       });
     });
   });
--- a/polyfill/intersection-observer-test.js
+++ b/polyfill/intersection-observer-test.js
@@ -73,7 +73,7 @@ describe('IntersectionObserver', function() {
       expect(io.root).to.be(document);
 
       io = new IntersectionObserver(noop, {root: rootEl});
-      expect(io.root).to.be(rootEl); 
+      expect(io.root).to.be(rootEl);
     });
 
 
@@ -1403,11 +1403,21 @@ describe('IntersectionObserver', function() {
 
       it('handles tracking iframe viewport', function(done) {
         iframe.style.height = '100px';
+        iframe.style.top = '100px';
+        iframeWin.scrollTo(0, 110);
         // {root:iframeDoc} means to track the iframe viewport irrespective of toplevel viewport
         var io = new IntersectionObserver(
           function (records) {
+
+            var intersectionRect = rect({
+              top: 0, // if root=null, then this would be 100.
+              left: 0,
+              height: 90,
+              width: bodyWidth,
+            })
             expect(records.length).to.be(1);
             expect(rect(records[0].rootBounds)).to.eql(getRootRect(iframeDoc));
+            expect(rect(records[0].intersectionRect)).to.eql(intersectionRect);
             done();
           },
           { root: iframeDoc }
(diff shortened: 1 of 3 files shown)


samouri:
open question: do we want to do feature detection for iframe viewport tracking and replace native with the polyfill in that case?

I'm leaning towards no and allowing consumers to detect it themselves. If they prefer the polyfill they can run delete InOb from window before running the polyfill.

polyfill/intersection-observer.js
```diff
@@ -131,8 +131,12 @@ function IntersectionObserver(callback, opt_options) {
     throw new Error('callback must be a function');
   }
 
-  if (options.root && options.root.nodeType != 1) {
-    throw new Error('root must be an Element');
+  if (
+    options.root &&
+    options.root.nodeType != 1 &&
+    options.root.nodeType !== 9
```
dvoytenko, line 137: Let's make both the same: either `!=` or `!==`.

polyfill/intersection-observer.js
```diff
@@ -131,8 +131,12 @@ function IntersectionObserver(callback, opt_options) {
...
+  ) {
+    throw new Error("root must be a Document or Element");
```
dvoytenko, line 139: I think the styleguide here uses single quotes.

polyfill/intersection-observer.js
```diff
@@ -395,7 +399,10 @@ IntersectionObserver.prototype._monitorIntersections = function(doc) {
   });
 
   // Also monitor the parent.
-  if (doc != (this.root && this.root.ownerDocument || document)) {
+  if (
+    doc != ((this.root && this.root.ownerDocument) || document) &&
+    !isDoc(this.root)
```
dvoytenko, line 404: Is this a sufficient check? What if the `root` is a document, but in the parent subframe?

polyfill/intersection-observer.js
```diff
@@ -618,7 +627,7 @@ IntersectionObserver.prototype._computeTargetAndRootIntersection =
  */
 IntersectionObserver.prototype._getRootRect = function() {
   var rootRect;
-  if (this.root) {
+  if (this.root && !isDoc(this.root)) {
```
dvoytenko, line 630: If we're doing it this way, then I think we need to change the code below:

```
var html = document.documentElement;
var body = document.body;
```

To something like:
```
const doc = isDoc(this.root) ? this.root : document;
var html = doc.documentElement;
var body = doc.body;
```

polyfill/intersection-observer.js
```diff
@@ -714,8 +723,11 @@ IntersectionObserver.prototype._rootIsInDom = function() {
  * @private
  */
 IntersectionObserver.prototype._rootContainsTarget = function(target) {
-  return containsDeep(this.root || document, target) &&
-    (!this.root || this.root.ownerDocument == target.ownerDocument);
+  const rootDoc = isDoc(this.root) ? this.root : this.root.ownerDocument;
+  return (
+    containsDeep(this.root || document, target) &&
```
dvoytenko, line 728: Should this be `this.root || rootDoc`?

dvoytenko (changes requested):

polyfill/intersection-observer.js
samouri: thats interesting! I hadn't thought about how in same-origin nested iframes, the `root`  could be the document of a parent iframe. check should probably be closer to:

```js
var rootDoc = (this.root && (this.root.ownerDocument || this.root)) || document;
if (doc != rootDoc) {
   ...
} 

```

polyfill/intersection-observer.js
samouri: whoops, my `prettier` is poorly configured for this repo. fixing

polyfill/intersection-observer.js
samouri: definitely! similar misunderstanding of potentially using a child or parent's document for same-origin cases

polyfill/intersection-observer.js
samouri: is there a difference? if `this.root` is not set, then `document` and `rootDoc` should be the same thing right?

polyfill/intersection-observer-test.js
```diff
@@ -1398,6 +1401,31 @@ describe('IntersectionObserver', function() {
...
+              iframeDoc.documentElement.clientHeight || iframeDoc.body.clientHeight;
+
+            expect(records.length).to.be(1);
+            expect(records[0].rootBounds.top).to.be(0);
+            expect(records[0].rootBounds.left).to.be(0);
+            expect(records[0].rootBounds.right).to.be(subviewportWidth);
+            expect(records[0].rootBounds.width).to.be(subviewportWidth);
+            expect(records[0].rootBounds.bottom).to.be(subviewportHeight);
+            expect(records[0].rootBounds.height).to.be(subviewportHeight);
+            done();
```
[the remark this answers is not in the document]
samouri: How can I simulate a situation where they would differ?
Does this mean I should create a local viewport larger than the outer viewport (causing scrollbars on outer viewport)

polyfill/intersection-observer.js
samouri: done

polyfill/intersection-observer.js
samouri: fixed!

polyfill/intersection-observer.js
samouri: made both `!=`, to match the precedent.

polyfill/intersection-observer-test.js
```diff
@@ -1398,6 +1401,31 @@ describe('IntersectionObserver', function() {
...
+        // {root:iframeDoc} means to track the iframe viewport irrespective of toplevel viewport
+        var io = new IntersectionObserver(
+          function (records) {
+            var subviewportWidth =
+              iframeDoc.documentElement.clientWidth || iframeDoc.body.clientWidth;
+            var subviewportHeight =
+              iframeDoc.documentElement.clientHeight || iframeDoc.body.clientHeight;
+
+            expect(records.length).to.be(1);
+            expect(records[0].rootBounds.top).to.be(0);
```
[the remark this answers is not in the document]
samouri: done!

polyfill/intersection-observer.js
samouri: updated

polyfill/intersection-observer-test.js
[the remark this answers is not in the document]
dvoytenko: I think there are tests like this already that scroll the page a little?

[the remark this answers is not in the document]
dvoytenko: E.g. see [this](https://github.com/w3c/IntersectionObserver/blob/f212b238bcda529bdda06e08786f7941d6613e40/polyfill/intersection-observer-test.js#L1122-L1130).

polyfill/intersection-observer.js
```diff
@@ -618,19 +625,20 @@ IntersectionObserver.prototype._computeTargetAndRootIntersection =
...
+    var html = doc.documentElement;
+    var body = doc.body;
     rootRect = {
       top: 0,
       left: 0,
       right: html.clientWidth || body.clientWidth,
       width: html.clientWidth || body.clientWidth,
       bottom: html.clientHeight || body.clientHeight,
-      height: html.clientHeight || body.clientHeight
+      height: html.clientHeight || body.clientHeight,
```
dvoytenko, line 641: Can't do this in IE :( No trailing commas. You might have to disable the prettier.

polyfill/intersection-observer.js
dvoytenko: Hmm. Yeah. You're right about this one. But I guess either way if you've already updated it.

dvoytenko:
I think this is really close now.

samouri:
Darn. I believe I found a webkit bug.

Essentially, Chrome and FF ensure that rootBounds when `root:document` are relative to the specified document (top is always 0).  Safari gives the top based on current scroll position within the iframe.

See: https://output.jsbin.com/jutegazuxe

polyfill/intersection-observer-test.js
```diff
@@ -3022,6 +3077,31 @@ describe('IntersectionObserver', function() {
...
+        iframeWin.scrollTo(0, 110);
+        // {root:iframeDoc} means to track the iframe viewport irrespective of toplevel viewport
+        var io = createObserver(
+          function (records) {
+            io.unobserve(iframeTargetEl1);
+            var intersectionRect = rect({
+              top: 0, // if root=null, then this would be 100.
+              left: 0,
+              height: 90,
+              width: bodyWidth,
```
samouri, line 3093: self-nit: trailing comma

samouri:
Current status:
- Using the polyfill: all tests pass in IE11, FF, Chrome, and Safari.
- Using native InOb:
  - FF: tests fail. I'm seeing the same behavior on `master` so I've ignored this.
  - Safari: A single newly introduced test fails. I've opened a webkit bug since I believe it is an error there: https://bugs.webkit.org/show_bug.cgi
  - Chrome: A _different_ newly introduced test fails. Chrome is allowing for cross-document tracking which is explicitly against spec right now.

polyfill/intersection-observer-test.js
```diff
@@ -1398,6 +1401,58 @@ describe('IntersectionObserver', function() {
...
+          { root: iframeDoc }
+        );
+
+        io.observe(iframeTargetEl1);
+      });
+
+      // Current spec indicates that cross-document tracking yields
+      // an essentially empty IntersectionObserverEntry.
+      // See: https://github.com/w3c/IntersectionObserver/issues/87
+      it('Does not track cross-document elements', function(done) {
```
dvoytenko, line 1433: nit: lowecase

polyfill/intersection-observer-test.js
```diff
@@ -3022,6 +3077,31 @@ describe('IntersectionObserver', function() {
...
+              left: 0,
+              height: 90,
+              width: bodyWidth
+            })
+            expect(records.length).to.be(1);
+            expect(rect(records[0].rootBounds)).to.eql(getRootRect(iframeDoc));
+            expect(rect(records[0].intersectionRect)).to.eql(intersectionRect);
+            done();
+          },
+          { root: iframeDoc }
```
dvoytenko, line 3100: I think we also need tests when `rootMargin` is involved.

polyfill/intersection-observer.js
```diff
@@ -493,11 +500,18 @@ IntersectionObserver.prototype._checkForIntersections = function() {
     var intersectionRect = rootIsInDom && rootContainsTarget &&
         this._computeTargetAndRootIntersection(target, targetRect, rootRect);
 
+    var rootBounds = null;
+    if (!this._rootContainsTarget(target)) {
+      rootBounds = getEmptyRect();
+    } else if (!crossOriginUpdater || this.root) {
+      rootBounds = rootRect
```
dvoytenko, line 507: semicolon

polyfill/intersection-observer.js
```diff
@@ -703,7 +718,9 @@ IntersectionObserver.prototype._hasCrossedThreshold =
  * @private
  */
 IntersectionObserver.prototype._rootIsInDom = function() {
-  return !this.root || containsDeep(document, this.root);
+  var rootDoc =
+    (this.root && (this.root.ownerDocument || this.root)) || document;
+  return !this.root || containsDeep(rootDoc, this.root);
```
dvoytenko, line 723: I think this is not quite right and possibly the old code is more correct. This check now only tests that a root is inside its parent doc. But the parent doc itself could be actually disconnected.

polyfill/intersection-observer.js
```diff
@@ -978,6 +999,10 @@ function getParentNode(node) {
   return parent;
 }
 
+function isDoc(node) {
```
dvoytenko, line 1002: add jsdoc

dvoytenko (changes requested):
Looks very good!

polyfill/intersection-observer.js
samouri: Thanks for catching this! Idk why I modified this - it was probably while trying to get the cross-document test working, but ended up handling the case in `_rootContainsTarget`

polyfill/intersection-observer-test.js
samouri: one for both cross-origin and same-origin? or is only same-origin fine?

dvoytenko: I think we need both same and cross-origin. But if that makes any difference, we probably need a pretty simple cross-origin and more detailed same-origin.

dvoytenko:
>   * Chrome: A _different_ newly introduced test fails. Chrome is allowing for cross-document tracking which is explicitly against spec right now.

Can we link the crbug issue here as well?

dvoytenko (approved):

polyfill/intersection-observer-test.js
samouri: Done

philipwalton (approved):

PR #466, opened 2020-11-16
iframed inob: top should be in reference to Window

samouri:
Does this test case demonstrate a bug?

Assuming an InOb within an iframe that is tracking the viewport (root=null), if the iframe is fixed position and `top:100`, then I'd expect the intersection within it at `y=0` to have an intersection rect with `y=100`.

cc @dvoytenko


samouri:
Closing. @dvoytenko filed the related bugs:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1677644
- https://bugs.chromium.org/p/chromium/issues/detail?id=1149713
- safari?

PR #485, opened 2021-10-21
Allow setting rootMargin and thresholds

### LINKED ISSUE #428: Allow dynamically changing rootMargin
In whatwg/html#5510 I'm trying to specify  in terms of IntersectionObserver as per discussion in whatwg/html#5236

The desired behavior for lazy-loading is to dynamically change the `rootMargin` in response to things like scrolling and network changes. See https://github.com/whatwg/html/issues/5408#issuecomment-623683336

Since `IntersectionObserver` only allows setting the `rootMargin` upon initialization, this would require one of

* Register a new `IntersectionObserver` and observe all the same elements as the old one, and get rid of the old one. This seems bad, but I assume is what a JS polyfill would have to do?
* Have the browser internally update the `rootMargin` even though the public API doesn't allow for it.

I think the second option seems better, but per the [extensible web manifesto](https://extensiblewebmanifesto.org/) we try to have web features be explained by lower-level APIs without having special magic that isn't accessible to web developers.

Are there issues with modifying the `rootMargin` after initialization?

zcorpan:
Fixes #428.

Implementation commitment:

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -218,8 +218,8 @@ use-cases. Per-{{observe()}} options could be provided in the future if the need
 interface IntersectionObserver {
 	constructor(IntersectionObserverCallback callback, optional IntersectionObserverInit options = {});
 	readonly attribute (Element or Document)? root;
-	readonly attribute DOMString rootMargin;
-	readonly attribute FrozenArray&lt;double&gt; thresholds;
+	         attribute DOMString rootMargin;
+	         attribute FrozenArray&lt;double&gt; thresholds;
 	undefined observe(Element target);
 	undefined unobserve(Element target);
 	undefined disconnect();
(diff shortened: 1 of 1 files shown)


szager-chromium:
As with modifying rootMargin, we need to specify the effect of modifying thresholds on previousThresholdIndex:

https://github.com/w3c/IntersectionObserver/pull/486#issuecomment-1086191513

domfarolino:
This thread was buried in my inbox and I got to it by doing some cleaning up. Out of curiosity, is there any progress here? @zcorpan Do you plan to continue this?

zcorpan:
@domfarolino if there's impl interest I can make a plan. :)

domfarolino:
I see. Maybe we can just ping @szager-chromium and @emilio, and I'm not sure for Apple? ....

PR #501, opened 2022-07-06
Call [cssom-view]'s "getting the bounding box" algorithm directly

### LINKED ISSUE #55: Don't call public API methods; instead call their backing algorithms

miketaylr:
Closes #55

(And remove the stray package.json for the now-gone polyfill, oops).

***
<a href="https://pr-preview.s3.amazonaws.com/miketaylr/IntersectionObserver/pull/501.html" title="Last updated on Jul 6, 2022, 2:36 PM UTC (59d22ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/501/eb168cb...miketaylr:59d22ba.html" title="Last updated on Jul 6, 2022, 2:36 PM UTC (59d22ba)">Diff</a>


miketaylr:
PTAL @emilio @szager-chromium

polyfill/package.json
```diff
@@ -1,23 +0,0 @@
-{
```
emilio, line 1: Drive-by, but sure.

emilio (approved):

Issue #161, opened 2016-10-04
Is "unit of related similar-origin browsing contexts" the right thing to use in the rootMargin / rootBounds restriction?


mstange:
There are two issues here that were brought up when I talked to members of our DOM team at Mozilla:
- The word "related" is superfluous because the root and the target will always be in the same tab.
- Why did you choose "similar-origin" over "same-origin"? Is there a use case for allowing access between cross-origin documents which are similar origin? Could the same information be obtained with existing APIs or does this allow new information leaks?

szager-chromium:
That phrase is standard spec-ese:

https://html.spec.whatwg.org/multipage/browsers.html#unit-of-related-similar-origin-browsing-contexts

Is there a better or more accurate way to say this?

ojanvafai:
I think we were going for whether this frame is in the same origin as the top-level frame since if it is, then you have access to all the information in the top-level frame even if there's a cross-origin frame wedged between you.

I think we could also define this as being in the same origin as the parent frame.

I'm not really sure what's best here, but I'm hopeful we can still change Chrome's behavior here without breaking content if there's something we should be doing differently.

szager-chromium:
If I remember correctly, the issue is whether there are any intervening frames between target and root that are from a different origin.  In that case, we don't want to report root bounds.

ojanvafai:
IIRC @bzbarsky had opinions on this. I don't have strong opinions on this. We should do whatever is most consistent with other things on the platform. No reason for IntersectionObserver to be a special snowflake.

zetafunction:
Given the replies, it sounds like we really just want same origin, right? I think that matches what Chromium actually implements already.

However, if we want to be more strict and disallow intermediate frames that are cross-origin, then things might be more interesting from a compatibility standpoint.

bzbarsky:
I'm not sure I have a strong opinion on this.  As @ojanvafai points out, if you're same-origin with the top window you can already do window.top.whatever to get whatever you want out of it...  So the real question is whether any information about those intervening frames is leaked.  If it's not, then it doesn't matter what their origins are, seems to me.

emilio:
This whole concept is gone from the HTML spec now, but this is still used in IntersectionObserver.

mystor:
cc @annevk who did the `BrowsingContextGroup` spec changes

annevk:
[...]

PR #511, opened 2023-08-30
Add Scroll Margin to Intersection Observer

### LINKED ISSUE #431: Ability to have automatic rootMargin for all nested scroll containers
_Originally posted by @zcorpan in https://github.com/whatwg/html/issues/5408#issuecomment-623981827_:

>> Scrolling vertically & horizontally for:
>> 
>> * Element scroll container scrolling
>
>So, I'm not sure how this would work. In particular, for the image carousel use case, using only the implicit root (which I think browser implementations do now) would mean that there is no threshold for the element scroll container case, so those images would only start loading after they are partially in view.
>
> There is likely a performance hit to observing all scrollable elements when lazy images are used. Is there a good way to make it "do what I want" without adding more API surface? Or is the web developer explicitly setting the root the best way to solve this?

Should `IntersectionObserver` API provide for a way to apply a `rootMargin` for all scroll containers, without needing to create multiple observers (and respond to styling changes causing scroll containers to appear/disappear)?

One use case here is lazy-loading for image carousels. The page has a scrollable `div` and many images with horizontal scrolling. To speed up page load time and reduce network usage, the images that aren't initially visible are annotated with `loading=lazy`.

The `loading=lazy` feature is defined in terms of `IntersectionObserver`, with root set to implicit root for all elements.

As specified and implemented right now, this causes the images in the image carousel to only start loading when they are at least partially visible -- because the `root` is the top-level viewport, not the scroll container.

I think ideally an image in this case would start loading when:

* the scroll container is "about to become visible" or is visible
* the image in the scroll container is "about to become visible" or is visible

I don't know how to solve this, but I'm hoping for something that results in a good user experience with the current web developer ergonomics of only adding `loading=lazy` to the right images.

cc @domfarolino @emilio @smfr

tcaptan-cr:
Closes #431

 * [ ] Modified Web platform tests ([link to pull request](https://chromium-review.googlesource.com/c/chromium/src/+/4740913))

Implementation commitment:

 * [ ] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=1469500)

***
<a href="https://pr-preview.s3.amazonaws.com/tcaptan-cr/IntersectionObserver/pull/511.html" title="Last updated on Sep 12, 2023, 10:07 PM UTC (41baa1f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/511/9b6715e...tcaptan-cr:41baa1f.html" title="Last updated on Sep 12, 2023, 10:07 PM UTC (41baa1f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -381,6 +381,8 @@ Percentages are resolved relative to the width of the undilated rectangle.
 
 Note: {{IntersectionObserver/scrollMargin}} affects the clipping of <a for="IntersectionObserver">target</a>
 by all scrollable ancestors up to and including the <a>intersection root</a>.
+Both the {{IntersectionObserver/scrollMargin}} and the {{IntersectionObserver/rootMargin}}
+are applied to a scrollable <a>intersection root's</a> rectangle.
 
 Note: <a>scrollport</a> intersection rectangles are not affected by
 <a>pinch zoom</a> and will report the unadjusted <a>viewport</a>, consistent with the
--- a/index.bs
+++ b/index.bs
@@ -286,7 +286,7 @@ interface IntersectionObserver {
 		constructor, the value of this attribute is "0px 0px 0px 0px".
 	: <dfn>scrollMargin</dfn>
 	::
-		Offsets are applied to <a>scrollable intersection rectangles</a>,
+		Offsets are applied to <a>scrollports</a> on the path from <a>intersection root</a> to <a for="IntersectionObserver">target</a>,
 		effectively growing or shrinking the box that is used to calculate intersections.
 		<b>These offsets are only applied when handling <a>same-origin-domain targets</a>;
 		for <a>cross-origin-domain targets</a> they are ignored.</b>
@@ -371,14 +371,8 @@ or failure:
 	append a duplicate of the second element to |tokens|.
 7. Return |tokens|.
 
-The <dfn for="IntersectionObserver">scrollable intersection rectangles</dfn>
-for an {{IntersectionObserver}}
-are the rectangles of scrollable containers on the path from the
-<a>intersection root</a> to <a for="IntersectionObserver">target</a>
-which could clip the target's rectangle.
-
-When calculating the <a>scrollable intersection rectangles</a> for
-a <a>same-origin-domain target</a>, the rectangle is then expanded
+When calculating a <a>scrollport</a> intersection rectangle for
+a <a>same-origin-domain target</a>, the rectangle is expanded
 according to the offsets in the {{IntersectionObserver}}’s {{[[scrollMargin]]}} slot
 in a manner similar to CSS's 'margin' property,
 with the four values indicating the amount the top, right, bottom, and left edges, respectively, are offset by,
@@ -388,7 +382,7 @@ Percentages are resolved relative to the width of the undilated rectangle.
 Note: {{IntersectionObserver/scrollMargin}} affects the clipping of <a for="IntersectionObserver">target</a>
 by all scrollable ancestors up to and including the <a>intersection root</a>.
 
-Note: <a>scrollable intersection rectangles</a> are not affected by
+Note: <a>scrollport</a> intersection rectangles are not affected by
 <a>pinch zoom</a> and will report the unadjusted <a>viewport</a>, consistent with the
 intent of pinch zooming (to act like a magnifying glass and NOT change layout.)
 
--- a/index.bs
+++ b/index.bs
@@ -501,6 +501,7 @@ The IntersectionObserverInit dictionary</h3>
 dictionary IntersectionObserverInit {
 	(Element or Document)?  root = null;
 	DOMString rootMargin = "0px";
+	DOMString scrollMargin = "";
 	(double or sequence&lt;double>) threshold = 0;
 };
 </pre>
@@ -517,6 +518,18 @@ dictionary IntersectionObserverInit {
 		each either an <a>absolute length</a> or a percentage.
 
 		<pre class="example" highlight=js>
+			"5px"                // all margins set to 5px
+			"5px 10px"           // top & bottom = 5px, right & left = 10px
+			"-10px 5px 8px"      // top = -10px, right & left = 5px, bottom = 8px
+			"-10px -5px 5px 8px" // top = -10px, right = -5px, bottom = 5px, left = 8px
+		</pre>
+	: <dfn>scrollMargin</dfn>
+	::
+		This is a string of 0-4 components,
+		each either an <a>absolute length</a> or a percentage.
+
+		<pre class="example" highlight=js>
+			""                   // no scroll margins
 			"5px"                // all margins set to 5px
 			"5px 10px"           // top & bottom = 5px, right & left = 10px
 			"-10px 5px 8px"      // top = -10px, right & left = 5px, bottom = 8px
--- a/index.bs
+++ b/index.bs
@@ -604,15 +604,20 @@ and an {{IntersectionObserverInit}} dictionary |options|, run these steps:
 	If a list is returned,
 	set |this|'s internal {{[[rootMargin]]}} slot to that.
 	Otherwise, <a>throw</a> a {{SyntaxError}} exception.
-4. Let |thresholds| be a list equal to
+4. Attempt to <a>parse a scroll margin</a>
+	from |options|.{{IntersectionObserverInit/scrollMargin}}.
+	If a list is returned,
+	set |this|'s internal {{[[scrollMargin]]}} slot to that.
+	Otherwise, <a>throw</a> a {{SyntaxError}} exception.
+5. Let |thresholds| be a list equal to
 	|options|.{{IntersectionObserverInit/threshold}}.
-5. If any value in |thresholds| is less than 0.0 or greater than
+6. If any value in |thresholds| is less than 0.0 or greater than
 	1.0, <a>throw</a> a {{RangeError}} exception.
-6. Sort |thresholds| in ascending order.
-7. If |thresholds| is empty, append <code>0</code> to |thresholds|.
-8. The {{IntersectionObserver/thresholds}} attribute getter will return
+7. Sort |thresholds| in ascending order.
+8. If |thresholds| is empty, append <code>0</code> to |thresholds|.
+9. The {{IntersectionObserver/thresholds}} attribute getter will return
 	this sorted |thresholds| list.
-9. Return |this|.
+10. Return |this|.
 
 <h4 id='observe-target-element'>Observe a target Element</h4>
 
--- a/index.bs
+++ b/index.bs
@@ -587,7 +587,9 @@ which are initialized to empty lists and an internal
 <dfn attribute for=IntersectionObserver>\[[callback]]</dfn> slot
 which is initialized by {{IntersectionObserver(callback, options)}}</a>.
 They also have an internal <dfn attribute for=IntersectionObserver>\[[rootMargin]]</dfn> slot
-which is a list of four pixel lengths or percentages.
+which is a list of four pixel lengths or percentages, and an internal
+<dfn attribute for=IntersectionObserver>\[[scrollMargin]]</dfn> slot
+which is either an empty list or a list of four pixel lengths or percentages.
 
 <h3 id='algorithms'>
 Algorithms</h2>
--- a/index.bs
+++ b/index.bs
@@ -713,9 +713,11 @@ run these steps:
 		of the {{document}}, and update |container| to be
 		the <a>browsing context container</a> of |container|.
 	2. Map |intersectionRect| to the coordinate space of |container|.
-	3. If |container| has a <a>content clip</a> or a css <a>clip-path</a> property,
+	3. If |container| is scrollable, apply the {{IntersectionObserver}}’s
+		{{[[scrollMargin]]}} to the |container|'s clip rect.
+	4. If |container| has a <a>content clip</a> or a css <a>clip-path</a> property,
 		update |intersectionRect| by applying |container|'s clip.
-	4. If |container| is the root element of a <a>browsing context</a>,
+	5. If |container| is the root element of a <a>browsing context</a>,
 		update |container| to be the <a>browsing context</a>'s {{document}};
 		otherwise, update |container| to be the <a>containing block</a>
 		of |container|.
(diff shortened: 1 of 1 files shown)


tcaptan-cr:
PTAL @emilio @zcorpan @smfr

marcoscaceres:
Also, @tcaptan-cr, you are committing to becoming co-Editor? That's ok... but just making sure you are 100% ok with taking on co-ownership. I'll need to announce that to the group.

tcaptan-cr:
> Thanks @tcaptan-cr... just to make this even easier to review, would you mind splitting out the Makefile into its own PR (and possibly even the xref bikeshed fixes if it's not a huge pain?)

I split the [makefile into this PR](https://github.com/w3c/IntersectionObserver/pull/512), and the [bikeshed fixes into this PR](https://github.com/w3c/IntersectionObserver/pull/513)

szager-chromium:
FYI, this change was reviewed in tcaptan's mirror:

https://github.com/tcaptan-cr/IntersectionObserver/pull/1

szager-chromium (approved):
lgtm

tcaptan-cr:
> Also, @tcaptan-cr, you are committing to becoming co-Editor? That's ok... but just making sure you are 100% ok with taking on co-ownership. I'll need to announce that to the working group.

Yes, I am 100% OK with taking on co-ownership.

index.bs
```diff
@@ -630,9 +673,11 @@ run these steps:
 		of the {{document}}, and update |container| to be
 		the <a>browsing context container</a> of |container|.
 	2. Map |intersectionRect| to the coordinate space of |container|.
-	3. If |container| has a <a>content clip</a> or a css <a>clip-path</a> property,
+	3. If |container| is scrollable, apply the {{IntersectionObserver}}’s
```
emilio, line 676: Would be great to have a more precise definition of `scrollable`, or a link to the relevant CSS spec. Otherwise whether this applies to `overflow: hidden` or not is a bit questionable, right?

emilio (approved):
Looks good to me, with that bit clarified. Thanks for doing this!

index.bs
tcaptan-cr: Good point. How about using CSS [scroll-container](https://www.w3.org/TR/css-overflow/#scroll-container)?

emilio: Sounds good, let's make sure that we test overflow: hidden then, and confirm that it does apply.

index.bs
```diff
@@ -673,7 +673,7 @@ run these steps:
 		of the {{document}}, and update |container| to be
 		the <a>browsing context container</a> of |container|.
 	2. Map |intersectionRect| to the coordinate space of |container|.
-	3. If |container| is scrollable, apply the {{IntersectionObserver}}’s
+	3. If |container| is a <a href="https://www.w3.org/TR/css-overflow/#scroll-container">scroll-container</a>, apply the {{IntersectionObserver}}’s
```
emilio, line 676: Pretty sure there's a better way to do this properly with bikeshed. `<a>scroll container</a>` might be enough?

index.bs
tcaptan-cr: Thanks Emilio!
The tests I added in my [PR for the wpt tests](https://chromium-review.googlesource.com/c/chromium/src/+/4740913) use `overflow: hidden`. For example the [scroll-margin.html](https://chromium-review.googlesource.com/c/chromium/src/+/4740913/9/third_party/blink/web_tests/external/wpt/intersection-observer/scroll-margin.html) test.
I confirmed these pass with my [prototype](https://chromium-review.googlesource.com/c/chromium/src/+/4740520).

emilio (changes requested):
Let's fix the link. Pretty sure there's a better way of doing this.

Same for the `margin` property link, which I hadn't noticed.

Also, can you rebase to remove the merge commit?

index.bs
tcaptan-cr: That worked. Thanks for the suggestion!

tcaptan-cr:
> Let's fix the link. Pretty sure there's a better way of doing this.
> 
> Same for the `margin` property link, which I hadn't noticed.
> 
> Also, can you rebase to remove the merge commit?

Done.

emilio (approved):

tcaptan-cr:
> lgtm

Thanks Stefan!

tcaptan-cr:
> Looks good to me, with that bit clarified. Thanks for doing this!

Thanks Emilio!

letitz:
Hi there,

Chromium web platform security reviewer here. I've been looking at the I2P for this change and have a few questions, mostly to clarify my understanding:

1. Why does this rely on `same-origin-domain`-ness of elements? I see the spec [mentions](https://w3c.github.io/IntersectionObserver/#intersection-observer-interface:~:text=Any%20target%20of%20an%20explicit%20root%20observer%20is%20also%20a%20same%2Dorigin%2Ddomain%20target%2C%20since%20the%20target%20must%20be%20in%20the%20same%20document%20as%20the%20intersection%20root.) that this only matters for implicit intersection observers, but I don't have context on why this matters (or why origin-domain was chosen over origin, the usual security boundary).

2. I've given this some thought through the lens of [xs-leaks](https://xsleaks.dev/docs/attacks/experiments/scroll-to-text-fragment/), and I believe this makes no material difference to that kind of attack, mostly because `scrollMargin` does not apply to iframes. This means that even if the intersection rectangle has been expanded by `scrollMargin`, it will be clipped to the iframe's viewport on the way up the tree, so a malicious iframe will not be able to determine more than "am I visible" (modulo rootMargin, but it only applies for same-origin-domain elements). Does this match your understanding?
  I guess the iframe could be contained in a scroll container itself, and `scrollMargin` would affect that? This seems related to my next question:

4. In https://github.com/w3c/IntersectionObserver/issues/431#issuecomment-1542502858 it was mentioned that `scrollMargin` should stop at cross-origin iframe boundaries. What was meant by that, and is that specified in this PR? I cannot find a reference.

5. I cannot find in the spec algorithms where it is specified that `scrollMargin` only applies to `same-origin-domain` elements. Is the plain english paragraph normative here?

szager-chromium:
This feature is modelled after the existing [`rootMargin`](https://w3c.github.io/IntersectionObserver/#dom-intersectionobserver-rootmargin) feature, and it has basically the same set of security/privacy concerns...

1. Only implicit-root observers are subject to these restrictions, because only implicit-root observers communicate geometry information that crosses iframe boundaries. Quoting from the spec:

"Any target of an explicit root observer is also a same-origin-domain target, since the target must be in the same document as the intersection root."

2. `scrollMargin` *does indeed* apply to the top-level scrolling viewport of the iframe! No distinction is made between the document-level scrolling viewport and a scrollable element. That is why `scrollMargin` is restricted to explicit-root observers and `same-origin-domain` implicit-root observers.

3. The comment in that spec issue is simply stating the requirement that `scrollMargin` cannot probe the geometry of the embedding page, in the same manner that `rootMargin`. This is commensurate with the above explanations.

4. Beginning at line 291 in `index.bs`, this PR adds:

"<b>These offsets are only applied when handling <a>same-origin-domain targets</a>;
		for <a>cross-origin-domain targets</a> they are ignored.</b>"

letitz:
Thanks for the answers!

> This feature is modelled after the existing [`rootMargin`](https://w3c.github.io/IntersectionObserver/#dom-intersectionobserver-rootmargin) feature, and it has basically the same set of security/privacy concerns...

Right, and I don't have context on those concerns, hence my questions :)

> 1. Only implicit-root observers are subject to these restrictions, because only implicit-root observers communicate geometry information that crosses iframe boundaries. Quoting from the spec:
> 
> "Any target of an explicit root observer is also a same-origin-domain target, since the target must be in the same document as the intersection root."

I see, and this is actually specified by step 2.2.2 of the [Run the Update Intersection Observations Steps algorithm](https://w3c.github.io/IntersectionObserver/#update-intersection-observations-algo).

I'm still left with my questions:

1. Why does this matter (for explicit root observers)?
2. Why choose same-origin-domain instead of same-origin? The latter is the web's usual security boundary.

> 2. `scrollMargin` _does indeed_ apply to the top-level scrolling viewport of the iframe! No distinction is made between the document-level scrolling viewport and a scrollable element. That is why `scrollMargin` is restricted to explicit-root observers and `same-origin-domain` implicit-root observers.

Oh, I see. I was looking at step 3.1 of the [compute the intersection algorithm](https://w3c.github.io/IntersectionObserver/#calculate-intersection-rect-algo), but step 3.3 also applies because an iframe is a scroll container?

> 3. The comment in that spec issue is simply stating the requirement that `scrollMargin` cannot probe the geometry of the embedding page, in the same way that `rootMargin` cannot. This is commensurate with the above explanations.

I see, this is what is mentioned in the [Privacy and Security section](https://w3c.github.io/IntersectionObserver/#privacy). Can you help me understand the attack here? How would the embeddee learn about its embedder's geometry using `rootMargin` or `scrollMargin`?

That section mentions that `IntersectionObserver` does not allow new methods of detecting intersection, only making existing hacks more usable. Is this "compatibility goal" what led to the choice of same-origin-domain instead of same-origin?

> 4. Beginning at line 291 in `index.bs`, this PR adds:
> 
> "These offsets are only applied when handling same-origin-domain targets; for cross-origin-domain targets they are ignored."

Right, this was what I referred to as "the plain english paragraph". IIUC this does not correspond to any step in the [compute the intersection algorithm](https://w3c.github.io/IntersectionObserver/#calculate-intersection-rect-algo)? I was expecting to find something there.

szager-chromium:
> > 1. Only implicit-root observers are subject to these restrictions, because only implicit-root observers communicate geometry information that crosses iframe boundaries. Quoting from the spec:
> > 
> > "Any target of an explicit root observer is also a same-origin-domain target, since the target must be in the same document as the intersection root."
> 
> I see, and this is actually specified by step 2.2.2 of the [Run the Update Intersection Observations Steps algorithm](https://w3c.github.io/IntersectionObserver/#update-intersection-observations-algo).
> 
> I'm still left with my questions:
> 
> 1. Why does this matter (for explicit root observers)?

Sorry, I don't understand your question. Why does *what* matter for explicit root observers? Explicit root observers can use `scrollMargin` without restriction; implicit root observers can use `scrollMargin` conditionally.

> 2. Why choose same-origin-domain instead of same-origin? The latter is the web's usual security boundary.

Here's the [previous discussion](https://github.com/w3c/IntersectionObserver/issues/161).

> > 2. `scrollMargin` _does indeed_ apply to the top-level scrolling viewport of the iframe! No distinction is made between the document-level scrolling viewport and a scrollable element. That is why `scrollMargin` is restricted to explicit-root observers and `same-origin-domain` implicit-root observers.
> 
> Oh, I see. I was looking at step 3.1 of the [compute the intersection algorithm](https://w3c.github.io/IntersectionObserver/#calculate-intersection-rect-algo), but step 3.3 also applies because an iframe is a scroll container?

Yes, that's right.

> > 3. The comment in that spec issue is simply stating the requirement that `scrollMargin` cannot probe the geometry of the embedding page, in the same way that `rootMargin` cannot. This is commensurate with the above explanations.
> 
> I see, this is what is mentioned in the [Privacy and Security section](https://w3c.github.io/IntersectionObserver/#privacy). Can you help me understand the attack here? How would the embeddee learn about its embedder's geometry using `rootMargin` or `scrollMargin`?

The general idea is that the attacker would create a bunch of IntersectionObservers, each with slightly different `rootMargin` and perhaps a range of `thresholds`; and that by analyzing the received notifications, it would be possible to deduce the size of the containing viewport. I unfortunately can't find a concrete exploit, but this came up during the early days of developing the API.

> > 4. Beginning at line 291 in `index.bs`, this PR adds:
> > 
> > "These offsets are only applied when handling same-origin-domain targets; for cross-origin-domain targets they are ignored."
> 
> Right, this was what I referred to as "the plain english paragraph". IIUC this does not correspond to any step in the [compute the intersection algorithm](https://w3c.github.io/IntersectionObserver/#calculate-intersection-rect-algo)? I was expecting to find something there.

Ah, that's a good point. There is an explanation of the same-origin-domain restriction on `rootMargin` which is reachable from the algorithm description; but the equivalent explanation for `scrollMargin` is not directly reachable from the algorithm description. I think this content, which is currently intermingled with the definition of `root intersection rect`, should be moved to the [definition of `scrollMargin`](https://w3c.github.io/IntersectionObserver/#dom-intersectionobserver-scrollmargin-slot):

```
When calculating a scrollport intersection rectangle for a same-origin-domain target, the rectangle is expanded according to the offsets in the IntersectionObserver’s [[scrollMargin]] slot in a manner similar to CSS’s margin property, with the four values indicating the amount the top, right, bottom, and left edges, respectively, are offset by, with positive lengths indicating an outward offset. Percentages are resolved relative to the width of the undilated rectangle.

NOTE: scrollMargin affects the clipping of target by all scrollable ancestors up to and including the intersection root. Both the scrollMargin and the rootMargin are applied to a scrollable intersection root’s rectangle.
```

letitz:
Thanks for the answers and pointers, this is super helpful.

Using same-origin-domain makes sense for this, though it would always be better to avoid introducing a new dependency on `document.domain`. If y'all would be happy with same-origin, I think it would be a nicer fit for today's web platform.

+1 to amending the spec to extend the same-origin-domain restriction from `rootMargin` to `scrollMargin` more explicitly.

I have one remaining clarification question: the intent of the same-origin-domain restriction is to prevent `scrollMargin` from applying to scroll containers (including iframe documents) that are cross-origin-domain with the observer, right? That is, if A embeds B embeds A, the grandchild A cannot learn about the geometry of B by using `scrollMargin`, even though it is same-origin with the top-level document. Is that correct?

progers:
I don't think there is any new information being exposed that is not already in rootMargin. Do you think there may be something?

> the intent of the same-origin-domain restriction is to prevent scrollMargin from applying to scroll containers (including iframe documents) that are cross-origin-domain with the observer, right?

Yes, this is my understanding.

letitz:
> I don't think there is any new information being exposed that is not already in rootMargin. Do you think there may be something?

I think I'm now convinced there isn't. Since `scrollMargin` applies to many more elements than `rootMargin`, I was worried there could be.

> > the intent of the same-origin-domain restriction is to prevent scrollMargin from applying to scroll containers (including iframe documents) that are cross-origin-domain with the observer, right?
> 
> Yes, this is my understanding.

Thanks for confirming.

zcorpan:
There were no bugs filed for Gecko or WebKit. Please keep that in the issue template in the future.

I've now filed:

- https://bugzilla.mozilla.org/show_bug.cgi?id=1860030
- https://bugs.webkit.org/show_bug.cgi?id=263370

PR #515, opened 2023-09-28
Clarify scrollMargin details

tcaptan-cr:
RE: ["I think this content, which is currently intermingled with the definition of root intersection rect, should be moved to the definition of scrollMargin"](https://github.com/w3c/IntersectionObserver/pull/511#issuecomment-1736315746)

***
<a href="https://pr-preview.s3.amazonaws.com/tcaptan-cr/IntersectionObserver/pull/515.html" title="Last updated on Sep 29, 2023, 6:43 AM UTC (a47c2fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/IntersectionObserver/515/f22fbf4...tcaptan-cr:a47c2fc.html" title="Last updated on Sep 29, 2023, 6:43 AM UTC (a47c2fc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -288,8 +288,6 @@ interface IntersectionObserver {
 	::
 		Offsets are applied to <a>scrollports</a> on the path from <a>intersection root</a> to <a for="IntersectionObserver">target</a>,
 		effectively growing or shrinking the clip rects used to calculate intersections.
-		<b>These offsets are only applied when handling <a>same-origin-domain targets</a>;
-		for <a>cross-origin-domain targets</a> they are ignored.</b>
 
 		On getting, return the result of serializing the elements of {{[[scrollMargin]]}}
 		space-separated, where pixel lengths serialize as the numeric value followed by "px",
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -298,6 +298,20 @@ interface IntersectionObserver {
 		passed to the {{IntersectionObserver}} constructor.  If no
 		{{IntersectionObserverInit/scrollMargin}} was passed to the {{IntersectionObserver}}
 		constructor, the value of this attribute is "0px 0px 0px 0px".
+
```
szager-chromium, line 301: Sorry, my suggestion was not quite correct. The goal is to make sure that all the language describing how `scrollMargin` is applied is reachable from the algorithm description for "Compute the Intersection of a Target Element and the Root". That algorithm description currently refers to the `[[scrollMargin]]` slot, as distinct from the `scrollMargin` argument to the IntersectionObserver constructor.

I think instead it would make sense to put these two paragraphs back where they were, but add a line like 'To &lt;dfn&gt;apply scroll margin to a scrollport&lt;/dfn&gt;' above it. You should also move lines 291-292 ("&lt;b&gt;These offsets are only applied...") to the same place; and then refer to this new definition in the algorithm description for "Compute the Intersection of a Target Element and the Root", instead of referring directly to the `[[scrollMargin]]` slot.

Hopefully this makes sense...


So I think these two paragraphs should be moved to the place where the `[[scrollMargin]]` slot is defined; and the text on lines 291-292 ("These offsets are only applied when handling...") should also be moved there.

tcaptan-cr: Let me know if this looks better.

szager-chromium (approved):
lgtm, thanks!

tcaptan-cr:
> lgtm, thanks!

Thank you Stefan!

PR #516, opened 2023-10-02
Fix thresholds definition to match browsers.

emilio:
https://github.com/web-platform-tests/wpt/commit/8c0296b50bf3974677e024c04d45b3ff8b0e801e tested that thresholds when provided empty is [0].

This is not in the spec, but is reasonable and I'm happy to change Firefox to match this, given both WebKit and Blink do that.

 * [ ] Modified Web platform tests: N/A, WPT already tests this.

Implementation commitment:

 * [ ] WebKit: N/A (already has this behavior)
 * [ ] Chromium: N/A (same)
 * [x] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=1855836)


tcaptan-cr (approved):
Looks good to me

